PR #4, opened 2016-01-12
Provider restructuring

tcoulter:
- Created a new provider API, `TestRPC.provider()`. This provider sends requests through the same interface that the server does -- as opposed to sending requests directly through the manager -- reducing the risk of bugs by having two interfaces.
- Added a configurable logging mechanism. @ryepdx, I'd love to see if this interface will work for you. `TestRPC.provider(logger)` and `TestRPC.server(logger)`.
- Added most of @ryepdx's tests, and wrote them so they went through web3. These tests don't cover all requests TestRPC supports, but it's a good swath. As always, they could definitely be made better.
- `$ npm test` now runs the tests.

@kumavis and @ryepdx, I'd love it if you gave this a look. Thanks!


lib/server.js
```diff
@@ -95,6 +97,92 @@ Server = {
...
+            result += chunk;
+            if (typeof encoding == "function") cb = encoding;
+            if (cb) cb();
+          },
+          end: function() {}
+        };
+
+        // Hack an express middleware function so jayson does all the JSON RPC niceties for us.
+        // Note: jayson wants the payload parsed (i.e., non-stringified)
+        middleware(request, response, function() {
```
tcoulter, line 161: This is the only part of this PR I see as being particularly controversial (so I'm pointing it out to you). `jayson` only provides an ExpressJS middleware function for using a server as a library. So in order to make the function act as a web3 provider, I've had to create fake http request and response objects. Not hard, but definitely worth commenting on. The bad part about this is it might cause issues if the expected interfaces change; the good part is that both the provider and the server send requests through the same request flow.

kumavis, line 161: hmm i think using `jayson` is causing you more woes than it solves
here is my json parsing server and a little extra
https://github.com/MetaMask/eth-rpc-wrapper/blob/master/server.js#L60-L97

tcoulter, line 161: It's possible. I'm not ready to rip it out yet, and our tests confirm this "hack" works. I'm open to swapping but that should be another PR.

My gut is jayson is doing some nice things for us regarding processing requests and error handling, but I've yet to confirm those beliefs.

tcoulter, line 161: @kumavis Everything alright otherwise? If so, I'm gonna :shipit:

tcoulter:
Going to merge if there's no other comments.

kumavis:
:rocket:

ghost:
Thanks for handing this, @kumavis and @tcoulter!

PR #10, opened 2016-02-11
Add support for eth_sendRawTransaction

kumavis:
With tests!
All tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ The RPC methods currently implemented are:
 * `eth_mining`
 * `eth_newBlockFilter`
 * `eth_sendTransaction`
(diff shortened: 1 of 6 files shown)


tcoulter:
@kumavis FYI: Fixed `eth_getTransactionCount` and `eth_getBalance` for "non-existent" requests, where they'd return `undefined` instead of `0x0`. Also found a nasty bug with `eth_getTransactionCount` which would call its callback twice, which I've included here. And I fixed some cors-headers stuff for @simondlr. Waiting for him to test this in his environment, then I'll merge.

tcoulter:
Got the green light from @simondlr, this fixes his issue. Merging.

PR #13, opened 2016-02-15
Added missing async to package.json

simondlr:


tcoulter:
How did this happen? My fault of course. Good catch.

PR #19, opened 2016-03-02
Added eth_sign

MrChico:
`sgn = web3.eth.sign(signer_address, msghash)` outputs a hex string to sgn with ECDSA r, s, and v concatenated.
One can regenerate the signer of the signature with:

```
utils = require('ethereumjs-util');
var r = sgn.slice(2, 66);
var s = sgn.slice(66,130);
var v = sgn.slice(130, 132);
```

for an explanation of the follwing two lines, see [this issue](https://github.com/ethereumjs/ethereumjs-util/issues/26)

```
var v_buf = new Buffer(v, 'hex');
v = utils.bufferToInt(v_buf)+27;

var pub = utils.ecrecover(new Buffer(msghash, 'hex'), v, new Buffer(r, 'hex'), new Buffer(s, 'hex'));
addr = utils.pubToAddress(pub).toString('hex')
addr = utils.addHexPrefix(addr);
```

addr and signer_address should be the same

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@ The RPC methods currently implemented are:
 * `eth_newFilter` (includes log/event filters)
 * `eth_sendTransaction`
 * `eth_sendRawTransaction`
+* `eth_sign`
 * `eth_uninstallFilter`
 * `web3_clientVersion`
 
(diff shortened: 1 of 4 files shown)


MrChico:
Actually, this isn't right. Returning v as a value between 0 and 1 is the way the go team does it, but that is a bug. Let me rework it

tcoulter:
Very cool stuff either way. Excited for the feature. Thanks!

PR #27, opened 2016-03-09
Allow specifiying the gas limit when instantiating a Web3 provider.

ghost:
This is another change Dapple needs in order to work with TestRPC.


axic:
Why do we have a block gas limit in testrpc? Shouldn't it be something really high?

ghost:
Derp. Issued PR from master, should have been from feature branch.

PR #28, opened 2016-03-09
Allow specifiying the gas limit when instantiating a Web3 provider.

ghost:
Last PR was from the wrong branch. 

This is another change Dapple needs in order to work with TestRPC.

Previously, @axic wrote:

> Why do we have a block gas limit in testrpc? Shouldn't it be something really high?


ghost:
@axic: Yeah, it should be something really high, or at least the same as the Homestead limit. There should also be a way to change the limit as needed.

PR #29, opened 2016-03-10
Expose engine's stop and start functions.

ghost:
Engine continues polling after code execution is done, resulting in Dapple hanging instead of exiting. There needs to be some way to stop block polling.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -188,6 +188,12 @@ Server = {
       },
       send: function() {
         throw new Error("Synchronous requests are not supported.");
+      },
+      start: function() {
+        engine.start();
+      },
+      stop: function() {
+        engine.stop();
       }
     };
   },


tcoulter:
Thanks @ryepdx. Quick question. How are you using the TestRPC within your application? i.e., are you using `TestRPC.server()` or `TestRPC.provider()`? I have tests that use both methods to instantiate the provider and neither have an issue with keeping the process open. 

Here's my guess:
- Either you're using `var server = TestRPC.server()`, and forgetting to call `server.close()` when you're done; OR, 
- You're using `var provider = TestRPC.provider()` and I'm stumped as to why that would keep the process open (i.e., its for the reasons you mentioned).

If it's not the first case, can I see the code that's using the provider? Is it open-source? I'd love to investigate this myself so I at least understand what's going on before accepting the PR - though I'm happy to accept once I do. That said, the provider-engine will soon be removing block polling by default, and there might be a way that we can get away without block polling, so I'd like to investigate that as well.

jmdfm:
@ryepdx Hey can you confirm if this PR is still valid, or if @tcoulter's suggestions are sufficient for your use case. Will close this PR in 7 days if we don't hear from you. Thanks!

tcoulter:
[This pull request](https://github.com/MetaMask/provider-engine/pull/81) should fix your issue once it's been merged.

tcoulter:
This has been fixed, and is now available in v2.2.4. You shouldn't have to stop the engine anymore, hence no need to expose it. Cheers!

PR #34, opened 2016-03-16
Log when contract creation fails

vasuman:
Or _errback_


jmdfm:
@vasuman I was just looking to merge this - is there a reason you closed?

vasuman:
Return value is empty when contract creation fails for reasons other than out-of-gas also

tcoulter:
In the `develop` branch of the TestRPC, all transactions will log when there's a failure. I've even included a test for it. This will be released when `develop` is merged into `master` and a new version of the TestRPC is released. Thanks!

PR #49, opened 2016-03-31
Fix eth_hashrate

pelle:
I’m guessing ethereumjs-util changed it’s intToHex function to add a
hex prefix.


tcoulter:
Merged by hand!

PR #52, opened 2016-04-03
Added evm_setTimestamp and evm_mineBlocks

Georgi87:
Many test scenarios have time constraints. For example off chain trading requires a third party to have enough time to respond to your action. In order to not have to wait in real-time, it makes sense to offer evm calls to manipulate timestamp and block number easily. I added this functionality with `evm_setTimestamp` and `evm_mineBlocks`. The timestamp is now static and predictable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # SYNOPSIS
 
-Ethereum client for testing and development. Uses ethereumjs to simulate full client behavior and make developing Ethereum applications much faster. Includes all popular RPC functions and features (like events) and can be run deterministically to make development a breeze. 
+Ethereum client for testing and development. Uses ethereumjs to simulate full client behavior and make developing Ethereum applications much faster. Includes all popular RPC functions and features (like events) and can be run deterministically to make development a breeze.
 
 # INSTALL
 
@@ -103,6 +103,8 @@ There’s also special non-standard methods that aren’t included within the or
 
 * `evm_snapshot` : Snapshot the state of the blockchain at its current place. Takes no parameters. Returns the integer id of the snapshot created.
 * `evm_revert` : Revert the state of the blockchain to a previous snapshot. Takes one parameter. Reverts to the snapshot id passed, or the latest snapshot.
+* `evm_setTimestamp` : Sets timestamp of current block. Takes one parameter (unix timestamp). This is useful, if your tests have time-constraints for challenge periods etc..
+* `evm_setBlockNumber` : Sets block-number of current block. Takes one parameter (block number). This is useful, if your tests have block-number-constraints for challenge periods etc..
 
 These methods are really powerful within automated testing frameworks. Example uses for these methods are:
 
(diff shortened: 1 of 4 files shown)


aakilfernandes:
Split evm_setTimestamp into a separate PR here https://github.com/ethereumjs/testrpc/pull/105

tcoulter:
Hey guys - sorry I haven't gotten to this. Been back up lately. Will hopefully have a chance to regroup soon. @johnmcdowall - is this something you can evaluate?

aakilfernandes:
https://github.com/ethereumjs/testrpc/pull/107

is what I believe turns out to be a better alternative. Problem with setTimestamp is the clock stops ticking, which is usually not what the user expects

jmdfm:
I can take a look, but it might be over my head at the moment!

tcoulter:
I agree with @aakilfernandes that `evm_increaseTime` is better than `evm_setTimestamp`. I've added `evm_increaseTime` to the master branch here: https://github.com/ethereumjs/testrpc/commit/498eb6e2df034e9855bcb11826135c76aaa88349

Along with that I added the request `evm_mine`. This mines a single block, instead of many, but can easily be looped.

@Georgi87 @area @jbaylina You've all shown that you're in favor of this PR. Will `evm_increaseTime` and `evm_mine` work for you? Is it important to you that timestamps are exact?

area:
For testing edge cases in a reliable manner, I think `setTimestamp` would be more valuable (e.g. check that at `timestamp`, a function does one thing, and at `timestamp+1` it does another).

tcoulter:
Getting time to stop moving is a tricky bit. For instance, what happens when you mine the next block - should the time move again? You can't (or shouldn't) mine two blocks with the same timestamp, so you eventually have to start time again. Are there cases where you need up-to-the-second granularity? I wrote a script recently that uses the `increaseTime` feature to jump between time ranges - i.e., I jumped 8 days to ensure I was past the 1 week check required by the code I was using (link here: https://www.youtube.com/watch?v=MLtsmnhiXCE). That was "good enough" for my case, and I can see tests being written that are "good enough" in a similar way. 

There has been a significant refactoring of the codebase recently, which allows for different block processing strategies. One of the future strategies is to allow block processing by request; i.e., instead of insta-mining, the developer specifies when a block should be mined. My gut instinct is that a better way to solve this problem is to add the block time to that strategy -- i.e., you specify the time of the block when you're requesting it to be mined. That way the TestRPC doesn't have to keep track of when the time has stopped and started, and instead simply mines a block based on the input provided.

Another option is to ensure you can't mine a block in the past; i.e., if the new block to be mined doesn't have a timestamp that is greater than the timestamp of the last block, error. This would require the developer to call `setTimestamp` all the time, but it would at least ensure the blockchain isn't 4 dimensional. :)

area:
I agree that in most scenarios tests akin to the one you describe will be 'good enough'.

Either of the two solutions you propose to allowing specific timestamps to be mined sound fine to me, which is still a feature I would like to have in the ideal world. I agree that most users would find time being stopped to be quite odd, and in the vast majority of cases what is currently implemented would be adequate and the least confusing.

tcoulter:
I'm going to close this ticket, and tack it onto #108, as I think that's the way to solve it (i.e., adding the timestamp when requesting a block be mined). Additionally, it'll error if the block is in the past. Thanks @area!

PR #59, opened 2016-04-05
Use eth-lightwallet instead of ethereumjs-wallet temporarily

tcoulter:
ethereumjs-wallet, at the moment, breaks Windows support.


tcoulter:
Whoops, this includes solc3 changes which adds homestead/delegatecall support as well.

axic:
@tcoulter released 0.5.2 which depends on the fixed version of scrypt.

PR #82, opened 2016-05-04
Update block gasLimit to be equal to Homestead

area:
Is there a reason this hasn't been done before I haven't run across? Without it, then there are contracts that work just fine on Morden/the actual network that can't be tested with `testrpc`...


tcoulter:
No reason, just lots of work. 

Merging. I'm going to change the gas limit on line 389 as well (I know, it shouldn't be in two places). Will push a new version after and make a comment here.

tcoulter:
v2.0.7 pushed to npm.

PR #104, opened 2016-06-16
Fix #103

### LINKED ISSUE #103: Incorrect Signatures
When calling `web3.eth.sign`, the `v` value returned is incorrect. It currently be either a `1b` or a `1c` while in Geth, it will only be a `00` or a `01`.
## Expected Behavior

`web3.eth.sign(<address>,  <dataToSign>).slice(130, 132)` should return either `00` or `01`.
## Current Behavior

`web3.eth.sign(<address>,  <dataToSign>).slice(130, 132)` returns either `1b` or `1c`.
## Possible Solution

Subtract the `v` value by `1b` (27) before it is returned.
## Steps to Reproduce (for bugs)
1. Call `web3.eth.sign` with whichever parameters.
2. Run `.slice(130,132)` to get the `v` value.
3. Notice that it is always `1b` or `1c` instead of the expected `00` or `01`.
4. Profit?
## Context

In my dapps which work with Geth, I have to explicitly add 27 to my `v` values, always. This behaviour breaks signatures because the 27 is already implicitly added.

For the record, I like this because it's more user friendly but unless the main Ethereum providers change it, this is inconsistent behaviour.
## Your Environment
- Version used: v2.0.8
- Operating System and version: Arch Linux

Denton-L:
This should fix #103.

close ethereumjs/ethereumjs-util#44


test/requests.js
```diff
@@ -199,7 +199,7 @@ var tests = function(web3) {
   	  sgn = utils.stripHexPrefix(sgn);
   		var r = new Buffer(sgn.slice(0, 64), 'hex');
   		var s = new Buffer(sgn.slice(64, 128), 'hex');
-  		var v = new Buffer(sgn.slice(128, 130), 'hex');
+  		var v = new Buffer((parseInt(sgn.slice(128, 130), 16) + 27).toString(16), 'hex');
```
Denton-L, line 202: Sorry, I've never used `Buffer` before. Please let me know if there's a more elegant way to do this.

tcoulter:
Hi there. Can you write me a test that breaks for which this code fixes? I don't know enough of the signing details to know what the effect this might have on the system. @wanderer or @kumavis, perhaps you'd know?

Denton-L:
Sorry, what do you mean by that?

Do you want me to write a test which would fail under this commit but would pass before this commit?

tcoulter:
The other way. Fail before, using this PR as the fix. The idea is that A) this will better communicate what this new code is supposed to fixing, and B) hopefully help me explain why the number 27 is important.

Denton-L:
Well, [this](https://github.com/Denton-L/testrpc/blob/761a29e244a94e4c5849e21af254fcd58108b2b2/test/requests.js#L202) should be it, unless you want me to go into a little more detail.

tcoulter:
Hmm, I didn't see that. 

Let me make sure I understand: The `v` value returned by the TestRPC is 27 greater than that returned by geth. Do you understand why this is? As well, can you verify and explain why this is the case for all signing requests? My bet is that if this is the issue, it needs to be fixed upstream as ecsign produces that v value. The 27 seems like a magic number otherwise.

Denton-L:
It's implemented upstream [here](https://github.com/ethereumjs/ethereumjs-util/blob/4ea5f4a9047abd25c46b7f99e47678f331d40756/index.js#L386).

Unfortunately, I don't really know enough about the finer details of this stuff to be making an argument about where the fix should be placed. I'll leave that up to someone more knowledgeable. 

EDIT: I believe that putting the change in ethereumjs-utils would have a greater potential of breaking a lot of projects that depend on it while placing it here would not break anything because it would just match the expected behaviour of Geth.

tcoulter:
Let me talk to @wanderer and see why the `27` is in ethereumjs-util. Very strange.

Denton-L:
@tcoulter, I was talking to @wanderer in ethereumjs/ethereumjs-util#44 and he believes that the change should be placed in this repository.

tcoulter:
Do you know if geth has the same results as cpp-ethereum and parity?

Denton-L:
I haven't tested cpp-ethereum yet but parity returns `00` and `01` as the `v` values.

tcoulter:
Alright @Denton-L, you have me sufficiently convinced. Thank you for your PR, and explaining to me why it was important. 👏

kumavis:
I think it should be resolved here ethereumjs/ethereumjs-util#45

tcoulter:
Since we have a test that ensures the correct signature I'm going to leave it merged. Whenever ethereumjs-util is upgraded the test will fail, signaling to us that we can remove the fix here.

kumavis:
After all this, geth is intending to change its behavior
https://github.com/ethereum/go-ethereum/issues/2053

Denton-L:
That was last referenced seven months ago, though. Also, other Ethereum clients are doing the same thing, I don't think this change is going to be happening any time soon?

PR #105, opened 2016-06-16
added evm_setTimestamp and test

aakilfernandes:


aakilfernandes:
After using this in a project, I think its better to go the .jump route suggested by EtherSim. Going to close this PR and go that route

Georgi87:
Is there now an equivalent to `.jump`?

PR #106, opened 2016-06-18
Fixing incorrect documentation

redsquirrel:
The `accounts` attribute of the object that can be provided to `.provider()` and `.server()` should not be a 2 dimensional array. It's an object with one required attribute: `balance`.


tcoulter:
This is incorrect. It should be an array of objects like you described. i.e., 
- `"accounts"`: `Array` of `Objects`. Each object should have a `balance` key and hex value. Can also supply `secretKey`.

I'll make this change then merge.

PR #107, opened 2016-06-20
Added provider.manager.increaseTime

aakilfernandes:


tcoulter:
The master branch has changed significantly, so I've tackled this myself. I don't want to maintain an interface other than the RPC interface, so I've made `increaseTime` into the RPC call `evm_increaseTime`. Code here: https://github.com/ethereumjs/testrpc/commit/498eb6e2df034e9855bcb11826135c76aaa88349

Thanks @aakilfernandes!

PR #109, opened 2016-06-23
decrease default balance

aakilfernandes:
The previous default exceeds the max that BigNumber.js can handle, making it impossible for use in unit tests. This decreases the default balance to 100 eth (100000000000000000000 wei)

### DIFF
--- a/lib/blockchain.js
+++ b/lib/blockchain.js
@@ -102,7 +102,7 @@ Blockchain.prototype.addAccount = function(opts, callback) {
   if (opts.balance) {
     account.balance = to.hex(opts.balance)
   } else {
-    account.balance = "0x00000000000000000ffffffffffffff";
+    account.balance = "0x0000000000000056bc75e2d63100000";
   }
 
   this.stateTrie.put(address, account.serialize(), function(err, result) {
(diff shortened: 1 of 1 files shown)


tcoulter:
But don't you want to be rich @aakilfernandes? ;) 

Good catch. Surprised this hasn't been an issue earlier. Also: BigNumber has a max number? Can you provide some example code?

aakilfernandes:
So there's no "max", but a max which can be represented correctly. I'm looking for specific documentation but this is what I'm referring to

screen shot 2016-06-23 at 12 26 41 pm

tcoulter:
If I did my math right, this is only 1000 Ether? Should be enough, but I'd like to point it out for the github record.

aakilfernandes:
Yup. Thats way too low and web3 should fix this. I opened up an issue here a while ago https://github.com/ethereum/web3.js/issues/448

jmdfm:
Shouldn't this issue be raised with BigNumber?

aakilfernandes:
@johnmcdowall web3.js is using a forked version of bignumber. After looking at the branch they're using, its not obvious why

jmdfm:
Ahhhh.... sigh

PR #112, opened 2016-06-28
fix tests to reflect smaller balance

aakilfernandes:
Fixes a test that broke when the default balance got lowered


tcoulter:
Thanks @aakilfernandes!

Issue #108, opened 2016-06-22
Feature request: Control over transaction inclusion in blocks


PeterBorah:
It would be useful to have fine-grained control over blocks and the transactions they include. This would allow you to test situations where multiple transactions are in the same block, the precise block number matters, or large numbers of blocks pass between transactions.
## Expected Behavior

Various implementations are possible, but for example:

You send `testrpc` the request to enter Manual Block mode. From that point, until you request it return to normal mode, it does not automatically advance blocks every time it receives a transaction. Instead, it queues the transactions it receives until it receives the command to advance a block, at which point it does so and includes any queued transactions.
## Current Behavior

`testrpc` currently advances 1 block per transaction.
## Context

In multi-user scenarios, blocks are often used as proxies for elapsed time. It is therefore necessary to be able to test situations that occur only at specific block heights.

Additionally, there may be situations where you want to specifically test multiple transactions in the same block. For instance, if you have an oracle which only wants to reward a user for updating it at most once per block.

tcoulter:
- [ ] As a feature of this block processing strategy, whenever a block is requested to be mined, you can optionally pass the timestamp of the block to accurately test time code that's based on the passage of time. One condition of this feature is that you shouldn't be able to mine a block in the past. See comments in #52 for more discussion.

tcoulter:
@PeterBorah Everything you requested above has been added to the `develop` branch, primarily written by @area. The flow goes like this:
1. Send `miner_stop` request to the TestRPC to stop default mining behavior (i.e., instamine).
2. Queue transactions; they won't be mined when queued.
3. Send `evm_mine` request when you'd like to advance a block. Note that queued transactions are ordered by nonce and gas limit, so if you have a set of transactions whose gas limits total more than the default block gas limit, only a portion of those transactions will be mined (filing up the block without going over).
4. To start normal instamining again, send `miner_start` request to the TestRPC. The `miner_start`/`miner_stop` mimic the requests from `geth`.

I still need to test that `evm_mine` works correctly with the new flow, especially when handling error cases. I'd like `evm_mine` to return errors like OOG, but I have to see what I can do when multiple transactions are queued. Perhaps if you're queueing transactions, you forfeit the right to errors in the response since you're instamining. Not sure.

I'll attempt to write a test for this now / tomorrow morning.

tcoulter:
[...]

PR #116, opened 2016-07-04
[WIP] Fallback provider

jmdfm:
Implement a Fallback provider so that contracts called on the specified blockchain are copied into the testrpc 
- [x] Fetch current block number from fallback chain on startup
- [x] Intercept calls to `call`, `sendTransaction` & `getData`
- [x] Requests are routed to the fallback provider by block number


tcoulter:
Alright, proof of concept is committed. This PoC includes:
- [x] Falling back to the fallback provider on storage requests, balance requests, code requests and calls.
- [x] Executing transactions on the main provider while using the state of the fallback provider, and saving the resulting state on the main provider (i.e., forking from the fallback chain).
- [x] Large refactoring to many core files to separate concerns; i.e., separate blockchain functions from API responses, separate fallback functions from normal blockchain functions, and separate blockchain functions from state management (i.e., TestRPC behavior, like insta-mining). With this refactoring, `blockchain_double.js` now expects transactions to be queued and these transactions will be processed when the next block is processed. This opens doors for further features like processing blocks on an interval like a real blockchain (#127) and allowing users to cherry pick when blocks are processed (#108). Both of these features could conditionally replace the default insta-mine behavior based on user input. Arguably the state manager is still doing too much right now; there's room for further refactoring there.

What still needs to be done:
- [x] Have the fallback provider lock in its effective block number; many requests to the fallback provider are still being called with "latest", which should provide some odd behavior if not respected. Also need to cook up a test for this. 
- [x] Have the TestRPC manage block numbers as if the TestRPC's blocks were added _on top of_ the fallback provider's blocks. This means if the fallback provider was branched at block 1099, the next block executed on the TestRPC will be referenced as 1100 (but it will be the first block in the TestRPC's block array). 
- [x] In conjunction with the point above, requests that contain an optional block number -- like `eth_getBalance` -- need to have that block number respected in relation to the block number that TestRPC branched from. So if a request comes in asking for the balance of an address at a block that existed before the TestRPC was fired up, it should be accurately directed to the fallback provider for processing.
- [x] The fallback API should probably accept any web3 provider. The command line should be able to specify a string which will be turned into an HTTP provider, but other fallback providers should be an option for compatibility.

tcoulter:
New task that needs to be done:
- [x] Get logs across the fork boundary, like nothing happened.

tcoulter:
Added various other tests and fixes, including making sure blocks actually contain the correct numbers relative to the fallback, and the genesis block of the main chain uses the fallback block's hash as its parent hash. At least one thing still outstanding, as shown in the tests:
- [x] Block numbers and block hashes _within_ contracts need to accurately represent the fallback provider. See here: http://solidity.readthedocs.io/en/latest/units-and-global-variables.html#special-variables-and-functions

I've already written a test which fails. This might be another doozy.

tcoulter:
Alright, that's done! So the only thing outstanding (that I know about) is this item from my first update above:

> In conjunction with the point above, requests that contain an optional block number -- like eth_getBalance -- need to have that block number respected in relation to the block number that TestRPC branched from. So if a request comes in asking for the balance of an address at a block that existed before the TestRPC was fired up, it should be accurately directed to the fallback provider for processing.

What I found while trying to implement this is that the TestRPC doesn't support `eth_getBalance`, `eth_getTransactionCount` and `eth_getCode` with a block number attached. This surprised me. The functionality should be fairly straightforward when implementing this for both the fallback and normal blockchain, given the current refactorings. I'm going to put that task on this PR:
- [x] Support block tags and block numbers for `eth_getBalance`, `eth_getTransactionCount` and `eth_getCode` normally as well as for the fallback provider.

tcoulter:
Another thought:
- [x] Test snapshotting and reverting still works correctly, as this PR messes with the stateTrie a lot. Should likely write some tests for this.

tcoulter:
Lots of progress. There's a bug I haven't been able to track down, though is reproducible:
- [x] When running the dapp store migrations against a (first) provider that falls back to another (second) provider, the migration will fail to deploy with an OOG error. If the first provider is standalone, however, and doesn't fall back, the migrations deploy correctly.

I'll have to investigate this tomorrow.

tcoulter:
Alright, that odd bug is fixed. Been working on it for two days only to realize it was due to a typo. The end result is that most things were using the correct trie, but the vm's state manager's cache was using a trie of its own. Oof.

tcoulter:
Another outstanding issue:
- [x] eth_getBlockByNumber/eth_getBlockByHash for blocks before the fork should return their transactions and uncles. Currently they don't. Not sure if this is a bug or a feature, or if perhaps we can release the fallback provider and add this feature later.
- [x] same with eth_getTransactionByHash and eth_getTransactionReceipt

tcoulter:
Calling this one done! 💃

Issue #118, opened 2016-07-11
event listening web3.js notation changes logging behavior for contract factory methods


SilentCicero:
## Expected Behavior

I should be able to use either event handling notations (`contractInstance.EventName().get` and `contractInstance.EventName()`) from the web3.js object and TestRPC, however, this does not happen. The result is different.
## Current Behavior

This works:

```
contractInstance.EventName({someFilterVar: someFIlterValue}, (err, result) => {});
```

This does not:

```
contractInstance.EventName({someFilterVar: someFilterValue}).get((err, result) => {});
```
## Possible Solution

I believe this could be due to the way the web3.js object interfaces with the RPC providers and thus TestRPC. I am getting this event problem with both the packed `.provider()` and the `http rpc -> server` providers.
## Steps to Reproduce (for bugs)
1. Setup a Basic contract with Events
2. Deploy the contract to TestRPC
3. Use a method
4. Listen for events with either notation (to see that one works and the other does not)
## Context

This is preventing me from building basic applications with TestRPC. I have to use the one notation, which is fine, however, events dont seem to be functioning properly, causing me a lot of trouble with very basic applications.
## Your Environment
- TestRPC 2.0.9 -- web3.js 15.3
- Ubuntu 64bit, 16gb Ram
## Side Note

I'm having a ton of trouble with events in general. They seem to not fire when they should, and can even crash my node if I listen for too many at a single time. This I believe to be related to issues: 
https://github.com/ethereumjs/testrpc/issues/113

tcoulter:
Hi @SilentCicero, I'm not sure this is an actual bug. I did a bunch of a heavy digging, and had a writeup all ready for you, only to realize that you're performing events in the wrong order. Your steps to reproduce are these:
1. Setup a Basic contract with Events
2. Deploy the contract to TestRPC
3. Use a method
4. Listen for events with either notation (to see that one works and the other does not)

Items 3 and 4 should be reversed - you should start listening before triggering the function that should fire the event.

The reasoning behind this is the TestRPC is so fast that it processes the transaction that occurred in step 3, puts it in a block and moves on to the next block before 4 can start listening. Essentially, what the TestRPC has done is exposed a race condition in your application. On geth, where there's a significant block time between sending a transaction and that transaction being processed, you're not likely to notice - i.e., the block likely won't be processed before you start listening. However, with the TestRPC, you're listening **after** you've already "made a sound", we'll say. Which doesn't make sense right? :) 

[...]

PR #119, opened 2016-07-12
Small Cleanup + Tests

tcoulter:
Little bit of cleanup. Also added tests for events in contrast with some folks saying events aren't working. For instance: https://github.com/ethereumjs/testrpc/issues/118

TL;DR for #118: TestRPC is so fast that it processes transactions and creates a new block before people's dapps start listening. This test is here to ensure that when things are done in the correct order (listen first, then trigger event) things work correctly.

### DIFF
--- a/test/events.js
+++ b/test/events.js
@@ -1,0 +1,148 @@
+var Web3 = require('web3');
+var TestRPC = require("../index.js");
+var assert = require('assert');
+var solc = require("solc");
+
+var source = "                     \
+contract EventTest {               \
+  event NumberEvent(uint number);   \
+                                   \
+  function triggerEvent(uint val) {        \
+    NumberEvent(val);        \
+  }                                \
+}"
+
+// Thanks solc. At least this works!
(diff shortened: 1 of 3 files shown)


kumavis:
looks good

test/requests.js
```diff
@@ -24,14 +24,12 @@ var contract = {
   position_of_value: "0x0000000000000000000000000000000000000000000000000000000000000000",
   expected_default_value: 5,
   call_data: {
-    gas: '0x2fefd8',
```
jmdfm, line 27: Why removing the gas here and below?

test/events.js
```diff
@@ -0,0 +1,148 @@
...
+
+            return cleanup(new Error("Received event that didn't have the correct value!"));
+          });
+        }, 500);
+      });
+    });
+  })
+};
+
+var logger = {
```
jmdfm, line 114: Could we maybe have the TestRPC check if a logger has been set, and if not substitute this null logger and save having to set it up all the time?

PR #128, opened 2016-07-24
overwrite fakeBlockchain

aakilfernandes:
etheruemjs-vm, unless given a blockchain object, uses a fake blockchain

https://github.com/ethereumjs/ethereumjs-vm/blob/master/lib/fakeBlockChain.js

This PR overwrites that to reference the blockchain maintained by testrpc


tcoulter:
Very interesting. Thanks @aakilfernandes. Will go out with the next npm release (let me know if you have immediate need).

Issue #130, opened 2016-07-26
Default gas limit for eth_sendTransaction


axic:
Currently the default limit seems to be `4712388` ([see here](https://github.com/ethereumjs/testrpc/blob/master/lib/blockchain_double.js#L378)).

Both the [JSON RPC "spec"](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_sendtransaction) and [geth](https://github.com/ethereum/go-ethereum/blob/667a386d87de64587061c2462773977e8a9dc98a/eth/api.go#L55) use `90000` as the default.

I think `testrpc` should too, so that anyone not supplying their own gas limit during a contract deployment will realise it, even before they get to deal with geth.

tcoulter:
Added to `develop` branch in [this commit](https://github.com/ethereumjs/testrpc/commit/fc0fb4cc728c1eb7df1745120151f9ef8a737584). CC @area. 

Closing.

PR #136, opened 2016-08-11
Sets default gas price to 0.02 szabo

uzyn:
to better mimic current livenet and testnet's gas price.

I understand that gas price is changeable via option, however I feel that a much larger default than the current 1 would probably be better even during development.

### DIFF
--- a/lib/blockchain_double.js
+++ b/lib/blockchain_double.js
@@ -53,7 +53,7 @@ BlockchainDouble = function(options) {
   this.wallet = hdkey.fromMasterSeed(bip39.mnemonicToSeed(this.mnemonic));
   this.wallet_hdpath = "m/44'/60'/0'/0/";
 
-  this.gasPriceVal = '1';
+  this.gasPriceVal = '0x4A817C800'; // 0.02 szabo
 
   if (options.debug == true) {
     this.vm.on('step', function(info){
--- a/test/requests.js
+++ b/test/requests.js
@@ -101,11 +101,11 @@ var tests = function(web3) {
   });
 
   describe("eth_gasPrice", function() {
-    it("should return gas price of 1", function(done) {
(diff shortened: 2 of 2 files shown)


chevdor (commented):
I would disagree with this changes for the following reason: changing that now means we need to keep up with the changes in price when it goes up or down.

Moreover, each user can use the --gasPrice (-g) option to provide the right value at a given time.

uzyn:
A more reasonable default is still better than 1 wei, which is multiple orders of magnitude smaller than livenet's value.

Regarding keeping up with changes, I am fine with either of the follow 2 approaches:
1. Keep up with live net's hardcoded value, or
2. Do not keep up, but have the default at a a more sensible value than 1 wei.

Noticed the PR is now having conflicts. Let me know if I should resubmit.

tcoulter:
It's a simple PR; not sure why there's conflicts. No need to resubmit. We'll take care of it if it's accepted.

@area and I have been talking about fixing #130 in the next big release (3.0). This feature is similar. @area - can I get your view on it?

area:
Yeah, I can't see any reason why this isn't a net benefit. Even if the price changes on the live net, we can certainly be a lot more correct than ten orders of magnitude wrong.

A knock-on effect that might want to be considered is increasing the default balance accounts start with, given accounts will be spending a lot more on gas now, but I could equally see that being left alone - 100 Ether is very reasonable.

tcoulter:
There was a conflict because `gasPriceVal` was moved to the state manager. 

I've merged this into the `develop` branch. Will close this ticket to show 3.0 milestone progress.

Regarding the default gas price, there's a discussion in #109. I'm on the fence about raising it or leaving it where it is. If 100 ETH is plenty, we should probably stick with that for now. However, if there are reports that people are running out of funds during unit tests than I'd definitely consider raising it.

PR #152, opened 2016-09-01
Fix getTransactionReceipt for nonexistent transaction

area:
If a receipt for a transaction that didn't exist was requested,
an error was thrown. Now it returns `null` like `geth`. I've modelled
the new code after the implementation of `eth_getTransactionByHash`,
which is clearly written to avoid the same issue there.

I've included a new test for this too (NB two tests fail, but
that's the case on master, too, so I've left those alone for 
now).


tcoulter:
Whoops, good call. Thanks @area!

PR #153, opened 2016-09-01
Fix looking at nonces of queued transactions

area:
Previously, no queued transactions would be found when looking for nonces due to buffers being objects, and so even if they contained the same data, they would never be found to be equal through a shallow `!=`. This isn't a concern for the current implementation, because transactions aren't queued for very long at all - they are mined as soon as they are received. This also means I can't really write a test that failed before this change - apologies for that!

However, I stumbled across this while working on adding a toggle for disabling mining a block as soon as a transaction is received. We are in a situation where we would like to test queuing a few transactions up, and then mine them together, and so being able to send transactions, not have a block mined, and then trigger a block with `evm_mine` would be useful. I've since heard rumours that such a feature is already in the pipeline though - is there any truth to that?

### DIFF
--- a/lib/blockchain_double.js
+++ b/lib/blockchain_double.js
@@ -192,18 +192,24 @@ BlockchainDouble.prototype.createBlock = function() {
 };
 
 BlockchainDouble.prototype.getQueuedNonce = function(address, callback) {
-  var nonce = 0;
+  var nonce = null;
 
   this.pending_transactions.forEach(function(tx) {
-    if (tx.from != address) return;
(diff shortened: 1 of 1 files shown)


tcoulter:
Hey @area. These changes are included in your other PR correct? If so, let's close this one. Otherwise, good catch!

area:
Yeah, it is, I was just trying to separate out the changes that were being made - even if the other PR had ended up being deemed unsuitable, I'd have tried to get this included anyway in case it tripped someone up in the future.

Issue #81, opened 2016-05-04
Intermittent "TypeError: Cannot read property 'pop' of undefined"


PeterBorah:
I run my truffle tests against testrpc many times a day, and maybe once every 20 runs or so, this happens:

```
/usr/local/lib/node_modules/ethereumjs-testrpc/node_modules/merkle-patricia-tree/baseTrie.js:359
  var lastNode = stack.pop()
                      ^

TypeError: Cannot read property 'pop' of undefined
    at CheckpointTrie.Trie._updateNode (/usr/local/lib/node_modules/ethereumjs-testrpc/node_modules/merkle-patricia-tree/baseTrie.js:359:23)
    at /usr/local/lib/node_modules/ethereumjs-testrpc/node_modules/merkle-patricia-tree/baseTrie.js:106:16
    at /usr/local/lib/node_modules/ethereumjs-testrpc/node_modules/merkle-patricia-tree/baseTrie.js:442:14
    at processNode (/usr/local/lib/node_modules/ethereumjs-testrpc/node_modules/merkle-patricia-tree/baseTrie.js:447:23)
    at /usr/local/lib/node_modules/ethereumjs-testrpc/node_modules/merkle-patricia-tree/baseTrie.js:485:11
    at /usr/local/lib/node_modules/ethereumjs-testrpc/node_modules/merkle-patricia-tree/baseTrie.js:179:7
    at /usr/local/lib/node_modules/ethereumjs-testrpc/node_modules/merkle-patricia-tree/util.js:75:7
    at /usr/local/lib/node_modules/ethereumjs-testrpc/node_modules/async/lib/async.js:52:16
    at /usr/local/lib/node_modules/ethereumjs-testrpc/node_modules/async/lib/async.js:269:32
    at /usr/local/lib/node_modules/ethereumjs-testrpc/node_modules/async/lib/async.js:44:16

```

I haven't noticed any patterns about when this happens; it seems random to me.

Schaeff:
Same here!

tcoulter:
I haven't been able to reproduce this, but it's an intermittent failures
somewhere in the bowels of EthereumJS (looks like the merkle tree). My
hunch is that when a transaction is being saved some array that's supposed
to exist doesn't -- unfortunately I have nothing more to go on. If one of
you can find the smallest example that exposes the issue, that'd be very
helpful in debugging what's going on!

[...]

PR #155, opened 2016-09-02
Add ability to toggle mining

area:
Firstly, **do not merge!**. This depends on #153, and moreover needs some discussion before getting merged.

As mentioned in #153, we've found ourselves want to make sure transactions are included in the same block for testing, which is not possible with `testrpc` yet because as soon as a transaction is received, it is mined into a block. ~~This PR is a possible implementation of a flag `--noMineAfterTransaction` that toggles this behaviour.~~

The block-building logic becomes more complicated with multiple possible transactions, and is where the majority of the changes in this PR are found. In summary, when a block is mined:

- Each transaction is added to the block in turn, starting with the first pending transaction. If the first pending transaction fails, the bad transaction is removed from the pool, then an error is thrown.
- If subsequent transactions cause the block to fail when added, they _might_ be okay, but the block is getting full due to the block gas limit, or they might be bad transactions themselves. An error is not thrown, the transaction is not yet removed from the pool, and the remaining transactions are tried in turn.
- In any of these cases, the state trie is reverted after the block is tested.
- Finally, the block is mined with the transactions that it was able to include in this greedy fashion, and the transactions successfully included are removed from the list of pending transactions.

This implementation doesn't currently break any existing behaviour (the tests that pass on master still pass here). Some tests should be written for this new functionality, however.

The code needs some cleanup (notably failing DRY quite badly), but I thought it prudent to get any other comments at this point that would be able to improve this new feature to the level needed to get it released.

### DIFF
--- a/bin/testrpc
+++ b/bin/testrpc
@@ -46,7 +46,6 @@ var options = {
   fork: argv.f || argv.fork || false,
   network_id: argv.i || argv.networkId,
   verbose: argv.v || argv.verbose,
-  no_mine_after_transaction: argv.n || argv.noMineAfterTransaction || false,
   logger: console
 }
 
--- a/bin/testrpc
+++ b/bin/testrpc
@@ -46,6 +46,7 @@ var options = {
   fork: argv.f || argv.fork || false,
   network_id: argv.i || argv.networkId,
   verbose: argv.v || argv.verbose,
+  no_mine_after_transaction: argv.n || argv.noMineAfterTransaction || false,
   logger: console
 }
 
--- a/lib/blockchain_double.js
+++ b/lib/blockchain_double.js
@@ -9,6 +9,7 @@ var Trie = require("merkle-patricia-tree");
 var Web3 = require("web3");
 var utils = require("ethereumjs-util");
 var async = require('async');
+var Heap = require("heap");
 
 function BlockchainDouble(options) {
   var self = this;
@@ -200,6 +201,7 @@ BlockchainDouble.prototype.getQueuedNonce = function(address, callback) {
     if (tx.from.toString('hex') != address.toString('hex')) return;
 
     var pending_nonce = to.number(tx.nonce);
+    console.log('found nonce', pending_nonce);
     //If this is the first queued nonce for this address we found,
     //or it's higher than the previous highest, note it.
     if (nonce===null || pending_nonce > nonce) {
@@ -223,11 +225,67 @@ BlockchainDouble.prototype.queueTransaction = function(tx) {
   this.pending_transactions.push(tx);
 };
 
+BlockchainDouble.prototype.SortByPriceAndNonce = function(){
+  //Sorts transactions like I believe geth does.
+  //See the description of 'SortByPriceAndNonce' at
+  //https://github.com/ethereum/go-ethereum/blob/290e851f57f5d27a1d5f0f7ad784c836e017c337/core/types/transaction.go
+  //
+  var self = this;
+  var sortedByNonce = {};
+  for (idx in self.pending_transactions){
+    var tx = self.pending_transactions[idx]
+    console.log(tx.from.toString('hex'));
+    if (!sortedByNonce[tx.from.toString('hex')]){
+      sortedByNonce[tx.from.toString('hex')] = [tx];
+    }else{
+      Array.prototype.push.apply(sortedByNonce[tx.from.toString('hex')], [tx]);
+    }
+  }
+  var priceSort = function(a,b){
+    return parseInt(a.gasPrice.toString('hex'),16)-parseInt(b.gasPrice.toString('hex'),16);
+  }
+  var nonceSort = function(a,b){
+    return parseInt(b.nonce.toString('hex'),16)-parseInt(a.nonce.toString('hex'),16);
+  }
+
+  //Now sort each address by nonce
+  for (address in sortedByNonce){
+    console.log(address);
+    console.log(sortedByNonce[address]);
+    Array.prototype.sort.apply(sortedByNonce[address], nonceSort)
+  }
+
+  //Initialise a heap, sorted by price, for the head transaction from each account.
+  var heap = new Heap(priceSort);
+  for (address in sortedByNonce){
+    heap.push(sortedByNonce[address][0]);
+    //Remove the transaction from sortedByNonce
+    sortedByNonce[address].splice(0,1);
+  }
+
+  //Now reorder our transactions. Compare the next transactions from each account, and choose
+  //the one with the highest gas price.
+  sorted_transactions = [];
+  while (heap.size()>0){
+    best = heap.pop();
+    console.log('popped with nonce', best.nonce.toString('hex'), ' and gas limit ', best.gasLimit.toString('hex'));
+    if (sortedByNonce[best.from.toString('hex')].length>0){
+      //Push on the next transaction from this account
+      heap.push(sortedByNonce[address][0]);
+      sortedByNonce[address].splice(0,1);
+    }
+    Array.prototype.push.apply(sorted_transactions, [best]);
+  }
+  self.pending_transactions = sorted_transactions;
+}
+
 BlockchainDouble.prototype.processNextBlock = function(callback) {
   var self = this;
   var block = this.createBlock();
   var successfullyAddedTransactions = [];
 
+  //First, sort our transactions like geth does
+  this.SortByPriceAndNonce();
   function tryTransactionIdx(txnumber){
     Array.prototype.push.apply(block.transactions, [self.pending_transactions[txnumber]]);
     self._checkpointTrie();
--- a/lib/blockchain_double.js
+++ b/lib/blockchain_double.js
@@ -198,10 +198,9 @@ BlockchainDouble.prototype.getQueuedNonce = function(address, callback) {
   this.pending_transactions.forEach(function(tx) {
     //tx.from and address are buffers, so cannot simply do
     //tx.from==address
-    if (tx.from.toString('hex') != address.toString('hex')) return;
+    if (to.hex(tx.from) != to.hex(address)) return;
 
     var pending_nonce = to.number(tx.nonce);
-    console.log('found nonce', pending_nonce);
     //If this is the first queued nonce for this address we found,
     //or it's higher than the previous highest, note it.
     if (nonce===null || pending_nonce > nonce) {
@@ -234,25 +233,22 @@ BlockchainDouble.prototype.SortByPriceAndNonce = function(){
   var sortedByNonce = {};
   for (idx in self.pending_transactions){
     var tx = self.pending_transactions[idx]
-    console.log(tx.from.toString('hex'));
-    if (!sortedByNonce[tx.from.toString('hex')]){
-      sortedByNonce[tx.from.toString('hex')] = [tx];
+    if (!sortedByNonce[to.hex(tx.from)]){
+      sortedByNonce[to.hex(tx.from)] = [tx];
     }else{
-      Array.prototype.push.apply(sortedByNonce[tx.from.toString('hex')], [tx]);
+      Array.prototype.push.apply(sortedByNonce[to.hex(tx.from)], [tx]);
     }
   }
   var priceSort = function(a,b){
-    return parseInt(a.gasPrice.toString('hex'),16)-parseInt(b.gasPrice.toString('hex'),16);
+    return parseInt(to.hex(b.gasPrice),16)-parseInt(to.hex(a.gasPrice),16);
   }
   var nonceSort = function(a,b){
-    return parseInt(b.nonce.toString('hex'),16)-parseInt(a.nonce.toString('hex'),16);
+    return parseInt(to.hex(a.nonce),16) - parseInt(to.hex(b.nonce),16)
   }
 
   //Now sort each address by nonce
   for (address in sortedByNonce){
-    console.log(address);
-    console.log(sortedByNonce[address]);
-    Array.prototype.sort.apply(sortedByNonce[address], nonceSort)
+    Array.prototype.sort.apply(sortedByNonce[address], [nonceSort])
   }
 
   //Initialise a heap, sorted by price, for the head transaction from each account.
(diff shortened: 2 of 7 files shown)


tcoulter:
Thanks @area, this sounds awesome! I likely won't be able to get to this until next week, but I'm very excited for it. Appreciate it!

area:
Summary of the changes since pushed:

* Early testing revealed that more useful functionality would be the ability to toggle auto-mining. I have therefore implemented `miner_start` and `miner_stop`, mirroring the `geth` calls when `geth` is started with `--rpcapi "miner"` flag, and removed the `--noMineAfterTransaction` flag.
* Queued transactions are sorted before they start to be added to a block - I have reimplemented [the ordering algorithm](https://github.com/ethereum/go-ethereum/blob/290e851f57f5d27a1d5f0f7ad784c836e017c337/core/types/transaction.go#L384) used by default by `geth` (though I don't know go, so it's possible I've overlooked some subtleties of the implementation).
* When the miner is started after being stopped, it will mine all currently pending transactions
* A side effect to be aware of: when the miner is stopped, transactions are not checked for validity, as they are simply placed into the transaction pool. They are only checked when mining occurs. So for example, if mining is stopped and the first transaction from an account arrives with nonce `0x2`, no error is thrown. If a transaction with nonce `0x1` is received before mining starts again, both transactions will be mined successfully. If not, an error will be thrown and the response to the `miner_start` call will be the same as the response to the `send_Transaction` call would have been had mining been enabled when the transaction was submitted.

As before, tests that pass on master pass here, but still no tests exist for this functionality, which will likely be my next port of call.

tcoulter:
Hi @area. I want to emphasize that this sounds -- and looks -- awesome. Since devcon is looming, and this is a big change, it's unlikely I'll be able to give it the attention it deserves until afterward. But I do want to communicate my support. Thanks a ton for working on this.

Speaking of devcon, will you or any of your team be there? I'd love to meet you guys in person if I can. Very much appreciate all your hard work.

area:
That's fine, I appreciate it's a big change, and so doesn't want to be rushed in. I've been testing with it, and it feels like it's more prone to #81, though I don't really have any hard evidence on that front. Maybe it provides a clue to tracking that bug down though?

I'm afraid I'm not going to be at DevCon, but @elenadimitrova will be!

tcoulter:
#81 is an issue deep in the bowels of ethereumjs-vm, possibly (probably?) caused by something that we're doing within the testrpc. Since it's not readily reproducible, it's been very hard to track down. There's a completely rewritten ethereumjs-vm on its way (not sure about the ETA), but this bug might have to live until then.

tcoulter:
Looking at this now. Few questions from reading your messages:

* I was looking to call different mining algorithms "strategies", as in the user can choose which mining strategy they prefer. The default, right now, is "instamine" -- i.e., instant mining. I noticed you add the flag `--rpcapi "miner"`. If possible, I'd like to change this flag to `--strategy`, and then have a good word that describes this strategy. Perhaps we can call it "request", to mean "mining by request"?

* In thinking about the above, my assumption was that the TestRPC would start with a strategy and then never change its strategy during execution. However, it sounds like `miner_start` and `miner_stop` would let users turn instant mining off and on again, is that correct? I'm not sure if one method is better than the other. How do you use it? My gut instinct is to turn these sets of features into three separate strategies that don't intersect (can't be turned off an on):

  1. Instamine.
  1. Mine by request (your strategy)
  1. Mine after a specific block time. 

  You've done significant amount of work for (ii) that (iii) would not be hard to implement, something I can follow up with if needed. :clap:

I need to look over the code and make comments (if I have any) but as far as I can tell you've done a great job, and I really appreciate this feature.

tcoulter:
If you are amenable to my strategies idea, I want to see if I can refactor your work into a separate strategy (I'm hoping I can; need to actually try it before I can say I'll be successful).

area:
I think there's a small misunderstanding in some of your comments - the `--rpcapi "miner"` is only referring to geth - this PR currently doesn't contain any new flags (it did, but I removed it). When `geth` is started with those flags, it responds to the `miner_start` and `miner_stop` commands over the RPC to start and stop mining new blocks, respectively. My reasoning here was to keep `testrpc` and `geth` as compatible as possible, so as many tests could be run against both as possible.

I removed the startup flag I originally added - which corresponds to the 'mine on request' strategy you propose - because it was actually quite difficult to use effectively. For example, when doing a 'truffle test', we had to keep sending RPC commands for the initial deployment to actually work. With the `miner_stop` and `miner_start` commands, we can have 'instamine' most of the time, but when we want to ensure transactions are included in the same block, we have the ability to do that too.

Is 'mine after a specific block time' not already implemented through the `--blocktime` flag?

tcoulter:
>  My reasoning here was to keep testrpc and geth as compatible as possible, so as many tests could be run against both as possible.

Ahhh. I was confused. This makes much more sense now, and is a good decision. 

> Is 'mine after a specific block time' not already implemented through the --blocktime flag?

Technically yes, but very naively so. It's really instamine with the addition of mining at regular intervals: It'll still mine when you send a transaction, but will also mine a block every n number of seconds if you'd like the block number to keep moving forward. With the addition of this feature, it'll still mine even when you've turned mining off with `miner_stop` (unless you protected against that; I need to check). 

Just for my edification and to have a record here, can you provide an example test (in comments) of how you're using this? Would be nice to refer people to an example if they have questions.

Also, let's make sure I'm not confused. A test using this feature would generally look something like this:

1. `miner_stop`
1. Queue transaction
1. Queue another
1. ... (repeat) ...
1. `miner_start`. Once started, it immediately mines a block with what has been queued.

Corollary: If I queue a transaction after `miner_start` has been called, it'll instamine, correct?

If I'm correct in the above, I'll go ahead and try to fix the `--blocktime` parameter from mining when mining is stopped. As well, I'll see if I can perform any refactoring while I'm there, perhaps break out `--blocktime` from instamining at the same time.

tcoulter:
Looking over the code, it looks like transactions get run an exorbitant amount of times. Apparently the amount of times it'll get run is called the [triangle number](http://math.stackexchange.com/questions/593318/factorial-but-with-addition/593320). Effectively, if there are five transactions queued, it'll try the first on its own; then the first and the second; then the first, second, and third; then first, second, third, and fourth; and finally the first, second, third, fourth and fifth. It comes out to: `n(n+1)/2` where `n` is the amount of queued transactions. 

I'm not necessarily sure if there's a way to get around this -- at least if we're trying to attempt transactions ahead of time. I think what we should do is not only sort transactions by price and nonce, but also have `processNextBlock` only grab `x` number of transactions in a single block, where the total gas limit of `x` transactions < block gas limit. If we did that, we wouldn't have to try the transactions ahead of time. I'm going to see if I can give this a shot, will let you know how this goes.

tcoulter:
I took a stab a the above and created a new pull request: https://github.com/ethereumjs/testrpc/pull/170

I couldn't quite figure out how to get those changes to show up here, or how you can compare them with github, so you may have to compare my changes against yours via the command line. The gist of the changes are: 

- Performed some stylistic changes
- Removed preprocessing of transactions, so transactions are only processed once
- Stopped checkpointing the trie - this will leave failed transactions on the chain like all other blockchains (i.e., geth, etc). This is possibly controversial, but it removes the need to preprocess transactions at all.
- Refactored block processing logic such that block processing now only tries to grab as many transactions as will fit in the block, based on the transactions' gas limit. This processing could be smarter to stuff transactions even fuller, but this is good for now.
- Refactored `no_mine_after_transaction` flag to be called `is_mining`, and gave it a `StateManager.startMining()` and a `StateManager.stopMining()` interface.
- Edited queuedTransactions.js so that the `before` steps were within the `describe` block (minor change)
- Added three more tests in mining.js to ensure block processing works like we expect.

If you're okay with these changes, I'll go ahead and merge. Look forward to your feedback, and as always: Thanks!

tcoulter:
Ah, still in the todo: 

- [ ] Make `--blocktime` prevent instamining and only mine on the interval.

area:
Not gone through the code yet, but first comments:

>Also, let's make sure I'm not confused. A test using this feature would generally look something like this:
>
>miner_stop
>Queue transaction
>Queue another
>... (repeat) ...
>miner_start. Once started, it immediately mines a block with what has been queued.

Yep, this is exactly how we're using it, and means that these tests can be run on both `geth` and `testrpc`.

> _Avoiding triangle numbers_

This approach means that if a user doesn't specify a gas limit for their transaction, my understanding is that only one transaction will ever be included in a block, because `testrpc` assigns the whole block size as the gas limit if it's not defined (done [here](https://github.com/ethereumjs/testrpc/blob/master/lib/statemanager.js#L222). Seems like a good time to hit #130, which solves this, and again brings `testrpc` behaviour closer to `geth`.

> Stopped checkpointing the trie - this will leave failed transactions on the chain like all other blockchains (i.e., geth, etc). This is possibly controversial, but it removes the need to preprocess transactions at all.

I'm not sure I understand this. Does this mean `testrpc` won't error when a bad transaction is sent any more?

tcoulter:
> Yep, this is exactly how we're using it, and means that these tests can be run on both geth and `testrpc`.

This is beautiful. 

> My understanding is that this approach means that if a user doesn't specify a gas limit for their transaction, only one transaction will ever be included in a block

You're correct, good catch.

> Seems like a good time to hit #130, which solves this, and again brings `testrpc` behaviour closer to `geth`.

Very good point. We'll fix #130 at the same time, and bump TestRPC up to 3.0. 

> I'm not sure I understand this. Does this mean testrpc won't error when a bad transaction is sent any more?

It will error, but it will leave erroneous transactions on the blockchain. The checkpointing and reverting during transaction processing was solely there to prevent the results of erroneous transactions from remaining (for instance, nonce and balance changes). Leaving the failed transactions on the chain will bring behavior closer to `geth`. The hard part about this is that there's no clear spec on which failure scenarios get kept, and which get discarded. However, from personal experience, it seems the transactions that get discarded are those that can be weeded out before block processing ever occurs:

* wrong nonce
* insufficient funds (gas limit * gas price > balance)
* gas limit under the base gas value of 21000
* gas limit over the block gas limit

ethereumjs-vm does a lot of this error checking for us, but there's no way to distinguish between a validation error and a runtime error. If we were to add those validation checks in (we already have one of them in my branch) then potentially we'll be much closer to acting like `geth` in these cases. To play out this example, consider solidity code with a `throw` in it. That transaction has a runtime error, but the transaction, the decrease in balance due to the cost of the transaction (i.e., consuming all gas available), and the nonce change stays on the blockchain. 

I did a quick google to make sure I wasn't off my rocker on this. It seems like this is the case, but I wouldn't be surprised if there was an error case or two I haven't mentioned that does not get kept on the blockchain. Perhaps I should go read the whitepaper.

http://ethereum.stackexchange.com/questions/2307/why-does-a-solidity-throw-consume-all-gas

tcoulter:
As I'm rereading all of this, I'm noticing your PR is turning into more than just a single PR: It's turning into the release strategy for TestRPC 3.0. I think this is a good distinction, especially with the following features/changes:

- `miner_start`/`miner_stop` - ability to queue transactions into a block
- smart block processing and transaction ordering
- `--blocktime` without instamine, making `testrpc` act like a real blockchain
- 90000 default gas limit
- error handling like geth

Given these changes, along with all the features already available, 3.0 becomes very powerful. You can write tests with transactions queued; you can use the `testrpc` without instamining and mine on an interval, much closer to real life; and the defaults for the `testrpc` are closer to real life as well, hardening people's code. 

I've gone ahead and merged my branch into the `develop` branch, which includes your changes here. I think we can finish up 3.0 fairly quickly as the majority of the work is done, we just need to tie up a couple loose ends (mainly error handling and tests for those errors, but the `--blocktime` without instamining as well). I'd like you to give the `develop` branch a try, if you can. Given the tests that I wrote in the `test/mining.js` file that's added, I **think** it'll work fine for you without changes to your tests. Assuming you're amenable to those changes, I'll go ahead and close the PR's and attempt to get 3.0 out the door within the next two weeks (I get married soon; I want to get it done before then!). Let me know if this works for you.

Again, thank you for all the work that you've done here. You've realized the "strategies" features I've been wanting in a great way, and I really like how it's turned out. Thanks again!

area:
Our tests that use this functionality work on `develop`, though when trying to test multiple files, each with a single `contract()` block, I get an error (I'm hoping that's due to something else on the `develop` branch that you already know about, though?

```

TypeError: Cannot read property 'find' of undefined
    at Cache._update (/<snip>/testrpc/node_modules/ethereumjs-vm/lib/cache.js:137:23)
    at /<snip>/testrpc/node_modules/ethereumjs-vm/lib/cache.js:76:12
    at /<snip>/testrpc/node_modules/ethereumjs-vm/lib/cache.js:46:5
    at /<snip>/testrpc/node_modules/merkle-patricia-tree/baseTrie.js:76:5
    at /<snip>/testrpc/node_modules/merkle-patricia-tree/baseTrie.js:442:14
    at /<snip>/testrpc/node_modules/async/lib/async.js:52:16
    at done (/<snip>/testrpc/node_modules/async/lib/async.js:246:17)
    at /<snip>/testrpc/node_modules/async/lib/async.js:44:16
    at Object.return (/<snip>/testrpc/node_modules/merkle-patricia-tree/baseTrie.js:461:9)
    at processNode (/<snip>/testrpc/node_modules/merkle-patricia-tree/baseTrie.js:284:30)
```

tcoulter:
The fact that that error's all the way down in the ethereumjs-vm cache isn't promising. Is there any way I can get a copy of this code? I realize it's likely proprietary/stealth/whatever. Happy to sign some agreement.

Looking at where the stack trace is erroring, this is super weird. It's like the cache's internal object to store cache data is missing (which, if it's a cache, it shouldn't never have a place to cache data). If you can't share the project, any chance you can pare it down and try to reproduce with a smaller test/contract set?

area:
Happily, I can pare it right down!

```
//In an empty directory
truffle init
cp ./test/metacoin.js ./test/metacoin2.js
truffle test
```

In this case, the first couple of tests in the second file pass. I assume this is because the tests are quicker than the ones we have in our project, and there's a 'fun' race condition to deal with here.

tcoulter:
Wow. What a bug. Super confusing. Thanks for the easy repro steps!

tcoulter:
Fixed on `develop`. I had forgotten: a few weeks ago, I refactored block processing to act like a state change. That is, it would checkpoint before every block so you can revert to right before that block was processed (think, pushing blocks onto the blockchain then popping them off again). During my changes to your code, I had remove the checkpoint required to do this.

To fix, I simply added that checkpoint back in. I didn't add the revert in however, which only triggered on transaction failure; failed transactions will now stay on the blockchain like they do on geth.

tcoulter:
Let me know if your tests pass. Very excited to get this feature in.

area:
Looks good here!

tcoulter:
Sweet! Then I'm going to close this as it's been merged into `develop`. Thanks again for all your help!

PR #165, opened 2016-09-22
Move options to yargs

chevdor:
- Add help option and move options to yargs
  `testrpc --help
  `
- Fix for the generation of a given number of accounts
  `testrpc -t 3` would generated 10 accounts, this is fixed
- Version bump

PR #166, opened 2016-09-22
Show balances

chevdor:
In short, this PR changes how the list of account is displayed. Instead of

```
Available Accounts
==================
(0) 0x90f8bf6a479f320ead074411a4b0e7944ea8c9c1
(1) 0xffcf8fdee72ac11b5c542428b35eef5769c409f0
(2) 0x22d491bde2303f2f43325b2108d26f1eaba1e32b
(3) 0xe11ba2b4d45eaed5996cd0823791e0c93114882d
(4) 0xd03ea8624c8c5987235048901fb614fdca89b117
```

you get

```
Available Accounts
==================
(0) 0x90f8bf6a479f320ead074411a4b0e7944ea8c9c1 - 100.00000000000001 ETH
(1) 0xffcf8fdee72ac11b5c542428b35eef5769c409f0 - 100.00000000000001 ETH
(2) 0x22d491bde2303f2f43325b2108d26f1eaba1e32b - 100.00000000000001 ETH
(3) 0xe11ba2b4d45eaed5996cd0823791e0c93114882d - 100.00000000000001 ETH
(4) 0xd03ea8624c8c5987235048901fb614fdca89b117 - 100.00000000000001 ETH
```

### DIFF
--- a/bin/testrpc
+++ b/bin/testrpc
@@ -1,5 +1,5 @@
 #!/usr/bin/env node
-var argv = require('yargs').argv;
+var yargs = require('yargs');
 var TestRPC = require('..');
 var pkg = require("../package.json");
 var util = require("ethereumjs-util");
@@ -7,106 +7,14 @@ var URL = require("url");
 var Web3 = require("web3");
 var web3 = new Web3(); // Used only for its BigNumber library.
 
-<<<<<<< HEAD
 yargs
-.usage('Usage: $0 [options...] ')
-// .command('start', 'Start the server')
-.option('port', {
-  describe: 'The port to serve',
-  alias: 'p',
-  type: 'string',   // for compativility with previous version
-  default: '8545'
-})
-.option('hostname', {
-  describe: 'Hostname to listen on. Defaults to Node\'s server.listen() default.',
-  alias: 'h',
-  type: 'string'
-})
-.option('debug', {
-  describe: 'Output VM opcodes for debugging',
-  type: 'boolean',
-  default: false
-})
-.option('deterministic', {
-  describe: 'Generate deterministic addresses based on a pre-defined mnemonic.',
-  alias: 'd',
-  type: 'boolean',
-  default: false
-})
-.option('seed', {
-  describe: 'Use arbitrary data to generate the HD wallet mnemonic to be used.',
-  alias: 's',
-  type: 'string',
-  default: 'TestRPC is awesome!'
-})
-.option('mnemonic', {
-  describe: 'Use a specific HD wallet mnemonic to generate initial addresses.',
-  alias: 'm',
-  type: 'string'
-})
-.option('total_accounts', {
-  describe: 'Number of accounts to generate at startup.',
-  alias: 't',
-  type: 'integer',
-  default: 10
-})
-.option('blocktime', {
-  describe: 'Specify blocktime in seconds for automatic mining. Default is 0 and no auto-mining.',
-  alias: 'b',
-  type: 'integer',
-  default: 0
-})
-.option('gasPrice', {
-  describe: 'Use a custom Gas Price (defaults to 1)',
-  alias: 'g',
-  type: 'integer',
-  default: 1
-})
-.option('gasLimit', {
-  describe: 'Use a custom Gas Limit (defaults to 0x47E7C4)',
-  alias: 'l',
-  type: 'integer',
-  default: 4712388 // 0x47E7C4
-})
-.option('accounts', {
-  describe: 'Specify the accounts to generate at startup. testrpc --account="<privatekey with 0x>,<balance in wei>" [--account="<privatekey>,balance"]',
-  alias: 'a',
-  type: 'string'
-})
-.option('fork', {
-  describe: 'Fork from another currently running Ethereum client at a given block. Input should be the HTTP location and port of the other client, e.g. http://localhost:8545. You can optionally specify the block to fork from using an @ sign: http://localhost:8545@1599200',
-  alias: 'f',
-  type: 'boolean',
-  default: false
-})
-.option('network_id', {
-  describe: 'The network id',
-  alias: 'i',
-  type: 'integer',
-  default: 1377
-})
-.option('verbose', {
-  describe: 'Tell me more!',
-  alias: 'v',
-  type: 'boolean',
-  default: false
-})
-.option('help', {
-  describe: 'Show some help',
-  type: 'boolean',
-  default: false
-})
-.option('version', {
-  describe: 'Show the version',
-  type: 'boolean',
-  default: false
+.option("unlock", {
+  type: "string",
+  alias: "u"
 });
 
 var argv = yargs.argv;
 
-
-=======
->>>>>>> parent of 3b3c1ed... Add help option (testrpc --help)
 function parseAccounts(accounts) {
   function splitAccount(account) {
     account = account.split(',')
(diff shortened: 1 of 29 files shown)


tcoulter:
Like I mentioned in a separate PR, this PR extends from all your other code changes, cascading the resolution of any fixes required. Would be better if there's a way you could submit fixes in the future without the other changes included. I know that in practice this can be hard, but in general it makes our life easier when merging and discussing PRs.

bin/testrpc
```diff
@@ -84,7 +188,8 @@ server.listen(options.port, options.hostname, function(err, state) {
   var addresses = Object.keys(accounts);
 
   addresses.forEach(function(address, index) {
-    console.log("(" + index + ") " + address);
+    var balanceInEth = util.bufferToInt(accounts[address].account.balance)*1e-18;
```
axic: I think this should use `ethereumjs-units` or the unit conversion of web3.js.

chevdor: It would lead to the same result but indeed, I could use:

```
web3.fromWei(val, 'ether)
```

Although I see the comment (https://github.com/chevdor/testrpc/blob/556b19e475462916dab1ba7474e0cfe52b0f20fb/bin/testrpc#L8)

```
// Used only for its BigNumber library.
```

which makes me think that web3 could be removed in the future

axic: There's still `ethereumjs-units` which works nicely.

chevdor: ok, what benefit do you think it brings?

axic: It doesn't uses `Number` which is prone to overflow (53 bits) and it does conversion in a tested, proven way.

chevdor: In general I agree. In this current case, I tested up to 1000000000 ETH (so 1000000000e18) and I get no overflow issue. So I am not sure if this is worth adding a new dependency.
@tcoulter what is your take on this one?

tcoulter: In general, I'd prefer the conversion library (or `web3`), especially if it removes the `.00000000000001` trailing fractional part (I'm assuming it does). That said, I have a couple concerns I wanted to address:
1. this feature is only valuable when you change the default amount via `--accounts`. It's definitely a nice visual indicator, but if you don't use `--accounts` then the default will always be 100 ETH.
2. I've added a `locked` emoji on `develop` (see below) that signals whether an account is locked (new feature). Because a lot of changes have happened on `develop`, we'll likely have to remake this PR, possibly making the same changes by hand. For instance, I want this PR to only focus on printing the balances without including the `yargs` or `--account` changes.
   screen shot 2016-10-05 at 8 07 52 am

For concern (1), this is really easy to get over. I just need someone like @axic or @area (thanks for letting me ping you so often) to tell me it's a good feature. If other people find it valuable I'm happy to concede.

For (2), I could probably do this on my own today. It won't be hard as you've done the legwork that I can just refer to. 

Regarding `web3`, I **believe** it is used in `./bin/testrpc` (don't quote me on this) because the TestRPC already has `web3` as a dependency, and that includes its own version of `BigNumber`, so there's no need to include third dependency when I could just use the `BigNumber` library provided by `web3`. 

All that said, I'd much prefer `web3.fromWei(...)` instead of our own math.

area: I don't believe we've ever used --accounts. Perhaps it could display only if the balance is not 100ETH?

chevdor: @tcoulter to your comments:
1. Indeed the feature makes more sense when using --account but also for new users who may wonder what is the default balance of the accounts. Once it is displayed, this is clear for everyone.
2. Sweet! 

I would add that I find the 100 ETH rather arbitrary. This is fine as default, but I would find rather convenient to have an option to override the default. While 100 ETH is way enough for regular testing, one may want to test a DAPP and how its display behave with big amounts of ETH without having to bother with using --account and especially having to provide the private key.

Indeed **web3 is already a dependency** but there is a comment in the code that seems to suggest it may not stay. If it does (can someone comment on that?), I would rather use the existing web3 dependency than adding an extra one.

Since I did not change the behavior/options of --account, I am sure I isolate the change from the yargs one.

tcoulter: > Indeed web3 is already a dependency but there is a comment in the code that seems to suggest it may not stay. 

No no - that comment was meant to suggest that it's currently only being used in _that file_ for the `BigNumber` library. It's used extensively elsewhere, and will remain a dependency.

tcoulter: Perhaps we should remove that comment. :)

chevdor: OK clear. So I will do the following unless someone shouts 😄 :
- remove the yargs changes from here (we have them in #165 anyway)
- use web3 for the conversion as @axic was suggesting
- remove the confusing comment (https://github.com/chevdor/testrpc/blob/556b19e475462916dab1ba7474e0cfe52b0f20fb/bin/testrpc#L8)

Other wishes?

tcoulter: Merge with `develop`. I'd put the balance to the right of the lock symbol, and include the dash (like you currently have). 

I'm going to release TestRPC 3.0 which includes changes on `develop`. We can release this feature as a point release afterward. Thanks!

tcoulter:
Just merged `develop` into `master`. Changelog here: https://github.com/ethereumjs/testrpc/pull/177

bin/testrpc
chevdor: Side note, web3 conversion do not truncate the decimals:
`web3.fromWei('1234123414345234523452334', 'ether')`
gives
`1234123.414345234523452334`

So I will limit the decimals as well.

chevdor:
Here is a sample output:

```
$ ./bin/testrpc -d -a 3 --secure --unlock 0x90f8bf6a479f320ead074411a4b0e7944ea8c9c1
EthereumJS TestRPC v3.0.0

Available Accounts
==================
(0) 0x90f8bf6a479f320ead074411a4b0e7944ea8c9c1 🔓  - 100 ETH
(1) 0xffcf8fdee72ac11b5c542428b35eef5769c409f0 🔒  - 100 ETH
(2) 0x22d491bde2303f2f43325b2108d26f1eaba1e32b 🔒  - 100 ETH
```

chevdor:
It seems that preferences between:

```
(0) 0x139f459f164a2c96114f2d08638c12b2c4a74037 🔓  - 100 ETH
(1) 0xffcf8fdee72ac11b5c542428b35eef5769c409f0 🔐  - 100 ETH
(2) 0x22d491bde2303f2f43325b2108d26f1eaba1e32b 🔐  - 100 ETH
```

and 

```
(0) 0x139f459f164a2c96114f2d08638c12b2c4a74037 - 100 ETH 🔓
(1) 0xffcf8fdee72ac11b5c542428b35eef5769c409f0 - 100 ETH 🔐
(2) 0x22d491bde2303f2f43325b2108d26f1eaba1e32b - 100 ETH 🔐
```

are split. It is easy to change if needed.

I changed the 'locked' icon because the unlock one (🔓) look very similar (to 🔒) and I did not find a better one for the unlock. Using the 🔐, we see clearly the difference between 🔐 and 🔓

The reason for adding the 🔓 in the first place is that we should not assume that users know what will be the default state. Having an icon for both states makes it clear. Not to mention how cool they look 😄

benjamincburns:
Sorry to say so, @chevdor, but even ignoring the conflicts caused by the way this project has been restructured, I think this PR is a bit cluttered with lots of commits. I'd happily accept a single-commit PR like this, however, which just included the minimal change to display the account balance when the account list is dumped in `cli.js`.

chevdor:
Why closing this PR? Do you have the features already or do you want a rebase that squash the commits into a single one ?

Issue #23, opened 2016-03-08
Accept a genesis.json file


flyswatter:
The `genesis.json` file format specifies an initial state for the blockchain, including balances of account numbers.

This could be very useful for development environments where a client has a wallet on it, and we want them to have a balance even on the simulated blockchain.

Lets clients operate in the development environment with nothing but the RPC address and an account that they've mined.

(I'm talking about Metamask here).

axic:
Would be great! I was thinking about hacking in a command line option just to provide a list of accounts, but this would be way better.

axic:
A newer, more complex version of genesis.json: https://github.com/ethereum/wiki/wiki/Ethereum-Chain-Spec-Format

aakilfernandes:
Would also be useful for dev testing purposes (include pre-deployed contracts in the genesis)

benjamincburns:
Seems that this would eliminate the need for the request in #265, amongst other things - though the effort involved might be substantially larger (especially since there's already a PR for #265).

benjamincburns:
Per PR #406, we'll want to also make it possible to save a genesis file when none is specified.

john-osullivan:
Is anyone trying to work on this?  I'm in the scenario described by @aakilfernandes , trying to test a pre-deployed contract.  Having to spin up a real chain each time is a pain!

zcstarr:
Going to just ping this here, I think this is still a good idea 👍 , is there any major blocker to this?

PR #167, opened 2016-09-22
Better account creation

chevdor:
This PR should only be applied after #165 (since #165 contains the fix for the number of accounts that are created.

In the examples below some of the addresses and amounts are shortened for the sake of readability.

So far, we could do:
`testrpc -t 7`
to get 7 accounts.

or
`testrpc --account="0xabc123000....000, 3300000..." --account="0xabc123000000.....00000001, 33000....."`
to get 2 specific accounts.

With this PR, it becomes possible to issue:
`testrpc --account="0xabc123000....000, 3300000..." --account="0xabc123000000.....00000001, 33000....." -t 5`
and get the best of both: the accounts you specified, as well as some automatically generated accounts up to the amount you wish.

```
Available Accounts
==================
(0) 0x85bd94190bf78f81fe15058545a6483c3d3c30fe - 330 ETH
(1) 0x36a3421f972f3370f2cbd1322c2c896409d9c894 - 330 ETH
(2) 0x22d491bde2303f2f43325b2108d26f1eaba1e32b - 100000 ETH
(3) 0xe11ba2b4d45eaed5996cd0823791e0c93114882d - 100000 ETH
(4) 0xd03ea8624c8c5987235048901fb614fdca89b117 - 100000 ETH
```

NOTE: This PR also raises the default balance of accounts from 100 ETH to 100k ETH as using tools such as truffle can exhaust an account pretty quick, especially when running tests. Also, why be happy with 100 when you can have 100k 😄 

Note that a call such as:
`testrpc --account="0xabc1230000...00, 330000..." --account="0xabc12300...000001, 330000..." --account="0xabc12300...00002, 330000..." -t 2`
will return:

```
Available Accounts
==================
(0) 0x85bd94190bf78f81fe15058545a6483c3d3c30fe - 330 ETH
(1) 0x36a3421f972f3370f2cbd1322c2c896409d9c894 - 330 ETH
(2) 0x50525a16b8d6aa19f59a01835ffaab9fb3f7652e - 330 ETH
```

So if accounts are provided, the total_accounts number is ignored if below the amount of specified accounts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,12 +21,14 @@ Having problems? Be sure to check out the [FAQ](https://github.com/ethereumjs/te
 ##### Command Line
 
 ```Bash
+$ testrpc --help
+
 $ testrpc <options>
 ```
 
 Options:
 
-* `-a` or `--accounts`: Specify the number of accounts to generate at startup.
+* `-a` or `--accounts`: Specify the list of accounts to generate at startup.
 * `-b` or `--blocktime`: Specify blocktime in seconds for automatic mining. Default is 0 and no auto-mining.
 * `-d` or `--deterministic`: Generate deterministic addresses based on a pre-defined mnemonic.
 * `-m` or `--mnemonic`: Use a specific HD wallet mnemonic to generate initial addresses.
--- a/bin/testrpc
+++ b/bin/testrpc
@@ -1,5 +1,5 @@
 #!/usr/bin/env node
-var argv = require('yargs').argv;
+var yargs =require('yargs'); 
 var TestRPC = require('..');
 var pkg = require("../package.json");
 var util = require("ethereumjs-util");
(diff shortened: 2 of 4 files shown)


tcoulter:
Like I mentioned in a separate PR, this PR extends from all your other code changes, cascading the resolution of any fixes required. Would be better if there's a way you could submit fixes in the future without the other changes included. I know that in practice this can be hard, but in general it makes our life easier when merging and discussing PRs.

tcoulter:
> NOTE: This PR also raises the default balance of accounts from 100 ETH to 100k ETH as using tools such as truffle can exhaust an account pretty quick, especially when running tests. Also, why be happy with 100 when you can have 100k

There's a reason we chose 100 ETH, actually. The reason is because when converted to a string and printed, it'll print out it's full value in Wei rather than printing out something like `1e+21`. Though I wasn't a huge fan of the decrease at first, this is a sensible default that looks good in applications while still giving enough Ether to do what you need. 

We had a slightly longer discussion about it here: https://github.com/ethereumjs/testrpc/pull/109

Definitely shoot me a conversation on gitter before changing defaults. We're very picky about them! :)

chevdor:
OK, no problem at all. I don´t have much argument to defend the increase
and this PR makes it easy to force one account with whatever amount.
I can restore the 100.

benjamincburns:
Closing this, as it's quite out of date, and I feel the mechanism outlined in #23 is a better, more standard, approach.

chevdor:
Sure #23 is way more flexible, also way less convenient. Why closing this? It seems to me that both PRs can coexist.

PR #170, opened 2016-09-27
start_mining/stop_mining, refactoring of #155

tcoulter:
I can't quite figure out how to compare my changes with the changes on #155. However, here's the gist:
- Performed some stylistic changes
- Removed preprocessing of transactions, so transactions are only processed once
- Stopped checkpointing the trie - this will leave failed transactions on the chain like all other blockchains (i.e., geth, etc). This is possibly controversial, but it removes the need to preprocess transactions at all.
- Refactored block processing logic such that block processing now only tries to grab as many transactions as will fit in the block, based on the transactions' gas limit. This processing could be smarter to stuff blocks even fuller, but this is good for now.
- Added a check when transactions are queued to ensure their gas limit doesn't exceed the block gas limit.
- Refactored `no_mine_after_transaction` flag to be called `is_mining`, and gave it a `StateManager.startMining()` and a `StateManager.stopMining()` interface.
- Edited queuedTransactions.js so that the `before` steps were within the `describe` block (minor change)
- Added three more tests in mining.js to ensure block processing works like we expect.

### DIFF
--- a/bin/testrpc
+++ b/bin/testrpc
@@ -46,7 +46,6 @@ var options = {
   fork: argv.f || argv.fork || false,
   network_id: argv.i || argv.networkId,
   verbose: argv.v || argv.verbose,
-  no_mine_after_transaction: argv.n || argv.noMineAfterTransaction || false,
   logger: console
 }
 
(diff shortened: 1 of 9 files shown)


tcoulter:
Closing; this has been merged into the `develop` branch.

PR #172, opened 2016-09-30
Fix157: Fix of the -t issue

chevdor:
-t 3 produces still 10 accounts
This PR fixes that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ $ testrpc <options>
 
 Options:
 
-* `-a` or `--accounts`: Specify the list of accounts to generate at startup.
+* `-a` or `--accounts`: Specify the number of accounts to generate at startup.
 * `-b` or `--blocktime`: Specify blocktime in seconds for automatic mining. Default is 0 and no auto-mining.
 * `-d` or `--deterministic`: Generate deterministic addresses based on a pre-defined mnemonic.
 * `-n` or `--secure`: Lock available accounts by default (good for third party transaction signing)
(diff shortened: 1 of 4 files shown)


tcoulter:
I'm confused after looking through the current README. `-t` isn't a valid parameter; the correct one is `-a`. Did you change that?

chevdor:
No I did not change (as far as I know) any meaning but I think there was confusion (ie changes) along the way (before my changes). The doc for -a or --account says (today):
`-a or --accounts: Specify the number of accounts to generate at startup.`

However, this is totally wrong. -a does not expect a number but a list of accounts (ie private key) + balance as mentioned a few lines below in the doc:
`testrpc --account="<privatekey>,balance" [--account="<privatekey>,balance"]`

I tested and realised that --account needs indeed the list of accounts. Calling -a with a number does not work.
My guess is that -t came out later to replace maybe an older version of -a that was taking a number.

chevdor:
@tcoulter Before working further on his one, we should clarify what option should do what between -a -t --account --accounts

benjamincburns:
Closing this one as well. As Tim says above, `-t` isn't a valid argument that I'm aware of, and the current documentation is clear as to how `-a` or `-accounts` is meant to be used.

chevdor:
@benjamincburns I am confused.
Can you please mention what is your take on what `-a` is doing?
If this is to pass the TOTAL amount of accounts, then we miss how to pass accounts, if this is to pass accounts, then we need a way to pass the total amount of accounts, thus the `-t` like in total.

Are you just closing all PRs without merging?

PR #173, opened 2016-09-30
Implement rpc_modules for `geth attach` compatibility.

tjade273:
Geth provides a nice console, with web3 provided, via the `geth attach` command. Currently, this command fails with the message 

> Fatal: Failed to start the JavaScript console: api modules: unable to retrieve modules

because testrpc does not implement the `rpc_modules` method. This PR fixes this, and now a geth console may be attached to a testrpc instance with 

```
geth attach http://localhost:8545
```


tcoulter:
Looks good. Thanks @tjade273!

PR #179, opened 2016-10-10
Feature/persistent

mhhf:
There are a lot of this left to do, but this is a first **draft** in case anyone want take a look.
Unfortunately I struggle with a non deterministic test failure which might be caused my memdown and a lot bugs look familiar to #81.
However, all tests work fine in isolation.
- this PR depends on a modified version of [ethereumjs-blockchain](https://github.com/nexusdev/mod-blockchain): 
  - the modification enables to submit a block on arbitrary block height without having its parent
    In the long run the ethereumjs-blockchain dependency can be omitted with including db storage directly into blockchain_double.js


tcoulter:
Awesome @mhhf, thanks for your work on this.

Quick heads up: I'm going to be away from the 13th to the 31st, which means this PR likely won't get merged until after I'm back. I'm sorry for the delay. If you think you can make this production ready by Wednesday, happy to jump on it, however, and do what I can to get it merged.

Regardless of timing, TestRPC 3.0 came out last week. Do you think you can remerge with `master` when you get a chance? Will be awesome to test this out.

Thanks!

mhhf:
I can make a sprint in the next 2 days: finalizing all features, merging the current master and clean everything up, however I don't think I will solve the problems caused by #81 and solving them is required for a production ready version :(

tcoulter:
We should hammer down on that bug then. I suppose this gives us more time. :)

mhhf:
I solved one source of non-deterministic bugs which also might be related to #81 - 
ethereumjs uses levelup/memdown combo which they always initialize with at path "":
In case there are many db's (e.g. multiple caching, multiple tries, multiple blockchains) they interfere with each other - e.g. if cache.flush is called some important data for some different actor might get removed. 
another weird behavior with memdown is that it has difficulties to store mixed type key values:
merkle-patricia-tree just stores Buffer values as keys, however if the db is used to store strings or ints as keys as well memdown might not find them, because it has difficulties to compare them which was also leading to errors.
However, after removing those two bug sources, there is still one non deterministic bug left which causes some tests to fail, in special across different tests with opening the db and closing it. 
But as for now, all tests work in isolation

mhhf:
from geth 1.5 upwards there is a new `eth_getRawTransaction` RPC API, which solves a major problem in state reconstruction. With this in place, I will remove the temporary hack of storing fake transactions. Also @tcoulter do you think we can pick up the work on this PR any time soon?

tcoulter:
@mhhf :+1: I'll have to look into that new request.

About this PR: apologies. I'm spread super thin right now. Perhaps this is something @johnmcdowall might want to look into? (warning: low level testrpc arch stuff)

Regardless, I'll see what I can do.

mhhf:
@tcoulter I hear you :) also have very limited resources right now. Did you had a change look on the code? Spend some more time to finishing it as far as I could and here is the outcome:

**TDLR**: [Code is working](https://travis-ci.org/nexusdev/testrpc) although messy, need to sync up with you on how to proceed.

Long version:
Although its working, the current architecture wasn't designed to work with persistency (e.g. interlinked object instances are passed everywhere instead of json).
Also I cannot say I fully understood the big-architecture-picture of testrpc and really struggled to debug certain aspect, which required me to refactor major parts of the code, therefore I might have put some significant technical debt inside and with this might also have broken some parts, not covered by tests.

1. I noticed that event/ log syncing is pretty slow with the current approach of chain-forking: if a filter is installed from genesis to latest block, testrpc tries to inspect each block for logs, and because they aren't there it syncs each block from the fallback blockchain. This is **verry** slow. What I currently do is forwarding the event filter to the forked-blockchain if the filter lies in its range, otherwise construct the result as usually. This might not be the best strategy, but its way faster then the previous one. However, coming up with some alternative strategy for log synching/ event filtering might be a better approach here.

2. I made extensive usage of `ethereumjs-blockchain` which I had to modify in order to support chain forking and address ownership faking. The capabilities of `ethereumjs-blockchain` are similar to the native `blockchain_double.js` so currently the blockchain logic is distributed between both. I suggest either to outsource much logic from `blockhain_double.js` into `ethereumjs-blockchain`. Or omitting the usage of it and putting all logic into `blockchain_double.js`. Some feedback would be great on this problem.

3. Its great that `eth_getRawTransaction` is implemented as a rpc command, but I think I'll propose to keep the current 'hacky' tx/fake.js implementation, since this allows us to pass wrongly signed transactions while still being able to store them persistently. In order to achieve this I had to make some very ugly etherumjs-tx hacks: see `statemanager.js:20`

4. I'm using testrpc as a dependency, therefore I didn't approaches the cli. An important definition would be having a flag for a persistent storage e.g.: `--db=/usr/root/db/`. Currently in-memory storage is used by default.

5. I could resolve the most, but there is still some non deterministic bugs which mostly are caused by dbs interference and asynchronously opening/ closing the db while the provider-engine is still working [see this](https://travis-ci.org/nexusdev/testrpc/builds/178883290) which was solved by putting setTimeout 100 on closing the server/provider.

Overall: Do you want to schedule a call and go over those points/ plan a merge?

tcoulter:
Hey @mhhf, can you resolve the conflicts on this one? Regarding Travis, I just added a config I found on the net - not attached to it. If yours is shorter and it works, lets use yours. However, can you bump it up to 6.9.1 as that's the latest LTS support? Thanks!

tcoulter:
Started a first pass at this. I'm very excited about this feature, but want to propose an alternate way of getting it merged.

There are many changes in this PR, making it hard to grok. As far as I can tell you have at least three general items you're trying to include: 

1. Implementing db persistence
2. Using ethereumjs-blockchain to manage blocks and block data
3. Fixing errors (added `var`'s, etc.) and making things better or more performant (log processing when forking mentioned above, which is a good catch). 

For my sake, is there any chance we can break this out into multiple PRs, each including the smallest change possible? For instance, we could have a process that looks something like this:

* First PR: Fix some bad code (adding `var`'s, etc.). (This could be multiple PRs)
* Second PR: Beef up log processing in forked mode
* Third PR: Add db persistence with current structure
* Fourth PR: refactor to use ethereumjs-blockchain

I realize I'm asking you to tediously go back over the work you've already done, but this would make it a lot easier for me to discuss aspects of each part without accepting the whole. For instance, I'm very much into db persistence, but I'm not yet convinced ethereumjs-blockchain integration is good at the moment (although I'm sure I'll change my mind later). I'd rather see we inject db persistence into the current structure and tackle the refactoring afterward.

As far as db persistence is concerned, I'd love to hear where some of the road blocks were. For instance, I see you've made some functions asynchronous because of it (makes sense). Also, I'd like to know if your current implementation supports snapshotting and reverting, and if that actually reverts the db on disk (looks like it does). 

I wouldn't be surprised if you tell me that we need ethereumjs-blockchain in order to use persistence, as perhaps it plugs into ethereumjs-vm's leveldb integration. If that's the case, let me ponder a bit as to how we should move forward. Regardless, pulling out the changes into separate PRs that we can individually (and sequentially) talk about would help to get this in more quickly.

Thanks @mhhf. Really looking forward to getting this feature in there.

mhhf:
@tcoulter I merged the current master.

About the task splitting it out in different PR's - 
I started implementing the persistency layer with including ethereumjs-blockchain as it uses leveldown + memdown combo to store all state, implementing persistency was as easy as exchanging the storage layer from memdown to leveldown. All other persistency was build upon ethereumjs-blockchain, so I splitting those tasks into different PR's wouldn't make much sense. I'd even argue to export even more testrpc logic to ethereumjs-blockchain to simplify the current architecture.

> Also, I'd like to know if your current implementation supports snapshotting and reverting

I don't see any reason why shouldn't be supported. Also I don't know if it revert the changes on the disk, we would need to make some experiments on this. Let me explain my general strategy during this PR: Levelup (leveldb) is used all over ethereumjs implementation, but with memdown (in memory storage) as default storage adapter. What I did is just exchange the storage adapter from memdown to leveldown (persistent storage adapter) in some places like ethereumjs-blockchain, ethereumjs-vm and its use of the trie. So if revert did erase the changes in memory, it will do this on the disc as well. 
Overall I tried to change testrpc's logic as less as possible.

I see how several PR's would help you with the review and also I see some places where this could be done - like the missing var statements or the log processing logic. However those changes are minimal and those cases are the only things I can think of, but I can devote a few hours to separate them into an external PR. The rest is quite atomic and interdependent, therefore I see splitting them into different PR's as unnecessary overhead.

I think the easiest way to both of us to save time is to schedule a call/ several calls and go over the code.

tcoulter:
@mhhf Just pinged you on gitter to set up a time. Thanks!

lib/interface.js
```diff
@@ -204,6 +222,10 @@ var Interface = {
       },
       send: function() {
         throw new Error("Synchronous requests are not supported.");
+      },
+      close: function (cb) {
+        engine.stop();
+        options.db.close(cb);
```
tcoulter, line 228: The need to call `engine.stop()` was removed long ago (I believe; let me know if you have another need for it other than to prevent the TestRPC from keeping the process open, which was fixed). As an aside, do you know what would happen to the database if `db.close()` wasn't called? Would data be corrupt or some data not saved?

I'm trying to avoid the need to call `.close()` on the provider at all since that is something easy to forget. Perhaps we can call necessary `.close()` calls on the db (and possibly the engine) in something like `process.on("beforeExit")` ([link](https://nodejs.org/api/process.html#process_event_beforeexit))? In that case the only time the db won't be closed is if a SIGKILL killed it, and even with the function above it wouldn't close anyway since the process would be immediately killed.

lib/statemanager.js
```diff
@@ -18,7 +17,15 @@ var to = require('./utils/to');
...
 var txhelper = require('./utils/txhelper');
 
-StateManager = function(options) {
+// TODO: verry ugly hack - discuss alternatives here.
+Transaction.prototype._getSenderAddress =
+Transaction.prototype.getSenderAddress;
+
+Transaction.prototype.getSenderAddress = function() {
+  return this._from;
+};
```
tcoulter: Is `FakeTransaction` not as fake as I think? I was under the impression it already did this.

lib/statemanager.js
```diff
@@ -211,14 +217,18 @@ StateManager.prototype.getCode = function(address, number, callback) {
 StateManager.prototype.queueRawTransaction = function(rawTx, callback) {
   var data = new Buffer(utils.stripHexPrefix(rawTx), 'hex');
 
-  var tx = new FakeTransaction(data);
+  var tx = new Transaction(data);
+  tx._getSenderAddress();
```
tcoulter: Is this line being used?

lib/statemanager.js
```diff
@@ -281,6 +291,23 @@ StateManager.prototype.queueTransaction = function(method, tx_params, callback)
...
+
+  if (tx_params.v != null) {
+    rawTx.v = utils.addHexPrefix(tx_params.v);
+  }
+
+  if (tx_params.from != null) {
+    rawTx.from = utils.addHexPrefix(tx_params.from);
+  }
+
+
```
tcoulter, line 302: This and the above (lines 239-231) just look like a refactoring. Is that correct?

lib/statemanager.js
tcoulter: If it's definitely not as fake as I think it is, and we need this change, we should put this into its own file and `require` that file everywhere.

lib/statemanager.js
```diff
@@ -497,6 +520,13 @@ StateManager.prototype.getLogs = function(filter, callback) {
 
   var logs = [];
   var current = fromBlock;
+  var toFetch = false;
+
+  if( this.blockchain instanceof ForkedBlockchain &&
+    fromBlock < this.blockchain.getEffectiveBlockNumber(this.blockchain.fork_block_number)) {
+    current = this.blockchain.getEffectiveBlockNumber(this.blockchain.fork_block_number) + 1;
+    toFetch = true;
+  }
```
tcoulter, line 521: The idea behind `ForkedBlockchain` is to abstract all the details into its own piece of code so that there wouldn't need to be any if statements like these. Is there any way to put this calculation into the `ForkedBlockchain` class instead? You might need to overwrite `getLogs` there.

lib/subproviders/geth_api_double.js
```diff
@@ -128,7 +126,7 @@ GethApiDouble.prototype.eth_getBlockByNumber = function(block_number, include_fu
       logsBloom: to.hex(block.header.bloom),
       transactionsRoot: to.hex(block.header.transactionsTrie),
       stateRoot: to.hex(block.header.stateRoot),
-      receiptRoot: to.hex(block.header.receiptTrie),
+      receiptsRoot: to.hex(block.header.receiptTrie),
```
tcoulter, line 129: Good catch.

lib/subproviders/reactiveblocktracker.js
```diff
@@ -8,8 +8,7 @@ module.exports = ReactiveBlockTracker;
 function ReactiveBlockTracker() {
   this.methods = {
     "eth_call": "before",
-    "eth_getStorageAt": "before",
-    "eth_getLogs": "before"
+    "eth_getStorageAt": "before"
```
tcoulter, line 11: You removed `eth_getLogs` from here. I remember it being needed to speed things up, if not provide the correct functionality. Are you sure it's no longer required? (I'm basically looking for surety; open to removing it if it's no longer required.)

lib/utils/forkedblockchain.js
```diff
@@ -2,8 +2,9 @@ var BlockchainDouble = require("../blockchain_double.js");
 var VM = require("ethereumjs-vm");
 var Account = require("ethereumjs-account");
 var Block = require("ethereumjs-block");
+var Blockchain = require('ethereumjs-blockchain');
 var Log = require("./log.js");
-var Receipt = require("./receipt.js");
+// var Receipt = require("./receipt.js");
```
tcoulter: Looks like this isn't needed. Remove if it.

lib/utils/forkedblockchain.js
```diff
@@ -73,8 +88,12 @@ ForkedBlockchain.prototype.initialize = function(accounts, callback) {
     self.web3.eth.getBlock(blockNumber, function(err, json) {
       if (err) return callback(err);
 
-      var block = self.createBlock();
+      var block = new Block();
       block.header.parentHash = utils.toBuffer(json.hash);
+      block.header.stateRoot = utils.toBuffer(json.stateRoot);
+      block.header.difficulty = utils.toBuffer('0x00');
+      block.header.gasLimit = utils.toBuffer(json.gasLimit);
+      block.header.number = utils.toBuffer(json.number + 1);
```
tcoulter, line 96: Anything here we can pull out and put in `BlockchainDouble.createBlock()` for consistency? Looks like `difficulty` might be the only one. For my edification, what was the reason to move away from `createBlock()` and instead use `new Block()`?

lib/utils/forkedblockchain.js
```diff
@@ -297,25 +321,8 @@ ForkedBlockchain.prototype.getTransactionReceipt = function(hash, callback) {
...
-        block: self.getBlock.bind(self, receipt_json.blockNumber)
-      }, function(err, result) {
-        if (err) return callback(err);
-
-        var logs = receipt_json.logs.map(function(log) {
-          return new Log(log);
-        });
-
-        var receipt = new Receipt(result.tx, result.block, logs, receipt_json.cumulativeGasUsed, receipt_json.contractAddress);
+    self.web3.eth.getTransactionReceipt(hash, callback);
```
tcoulter, line 327: This blows my mind that we can get rid of all this! The tests pass. Good catch!

lib/utils/forkedblockchain.js
```diff
@@ -410,13 +417,8 @@ ForkedBlockchain.prototype.fetchNonceFromFallback = function(address, block_numb
   });
 }
 
-ForkedBlockchain.prototype.getHeight = function() {
-  return to.number(this.fork_block_number) + this.blocks.length;
-};
-
 ForkedBlockchain.prototype.getRelativeBlockNumber = function(number) {
-  number = this.getEffectiveBlockNumber(number);
-  return number - to.number(this.fork_block_number) - 1;
```
tcoulter, line 419: Hmm, there was a reason I did this calculation, but I don't remember now exactly. If `getRelativeBlockNumber` and `getEffectiveBlockNumber` are synonymous, shouldn't we remove one or the other?

In my head, `relative` meant relative to our own block store: i.e., 0-based index in memory, which is needed to know if this is our block or the forked block. `effective` meant to translate things like "latest" to a real number. Have you merged the two?

lib/utils/forkedblockchain.js
```diff
@@ -438,8 +440,7 @@ ForkedBlockchain.prototype.getBlockLogs = function(number, callback) {
 
   number = this.getEffectiveBlockNumber(number);
   var relative = this.getRelativeBlockNumber(number);
-
-  if (relative < 0) {
+  if (relative <= parseInt(this.fork_block_number)) {
```
tcoulter, line 441: Looks like this might answer my above question, as the answer might be "you don't need `getRelativeBlockNumber` anymore".

lib/utils/receipt.js
```diff
@@ -12,7 +12,7 @@ function Receipt(tx, block, logs, gasUsed, contractAddress) {
   for (var i = 0; i < block.transactions.length; i++) {
     var current = block.transactions[i];
     if (current.hash() == tx.hash()) {
-      this.transactionIndex = index;
+      this.transactionIndex = i;
```
tcoulter, line 15: Hmm, how did that ever work...

lib/utils/txhelper.js
```diff
@@ -28,6 +28,7 @@ module.exports = {
       gasPrice: utils.toBuffer("0x" + json.gasPrice.toString(16)),
       data: utils.toBuffer(json.input)
     });
+    tx._hash = json.hash;
```
tcoulter: I don't generally like variables prefixed with `_` that are set outside of the class file/object file itself. Can you tell me why this one is here?

package.json
```diff
@@ -18,25 +18,30 @@
...
+    "leveldown": "~1.5.0",
+    "levelup": "~1.3.2",
     "heap": "~0.2.6",
     "merkle-patricia-tree": "~2.1.2",
     "seedrandom": "~2.4.2",
     "shelljs": "~0.6.0",
     "solc": "0.4.6",
     "web3": "~0.16.0",
-    "web3-provider-engine": "~8.1.0",
+    "web3-provider-engine": "git+https://github.com/nexusdev/provider-engine.git",
```
tcoulter: I can't rely on these github repositories for a final release. Let's see if we can get your changes merged upstream in provider-engine and ethereumjs-blockchain merged upstream.

package.json
```diff
@@ -18,25 +18,30 @@
...
     "yargs": "~3.29.0"
   },
   "devDependencies": {
     "eslint": "~3.0.1",
     "eslint-config-standard": "~5.3.5",
     "eslint-plugin-standard": "~1.3.3",
-    "mocha": "~2.2.5"
+    "mocha": "~2.2.5",
+    "solc": "~0.3.0-1",
```
tcoulter, line 43: We should probably update `solc` here.

test/accounts.js
```diff
@@ -3,29 +3,32 @@ var TestRPC = require("../index.js");
...
-    }));
+    });
+    web3.setProvider(provider);
 
     web3.eth.getAccounts(function(err, accounts) {
       if (err) return done(err);
 
       assert(accounts[0].toLowerCase(), expected_address.toLowerCase());
-      done();
+      provider.close(done);
```
tcoulter: Again, we *might* be able to get around this which would remove the need for any for changes in many of these test files.

tcoulter:
Alright, added a bunch of comments (I should have probably started a review; sorry for all the emails). Thanks again @mhhf!

lib/interface.js
mhhf: huh? you are right. It used to crash frequently while testing persistency because of frequent db opening/ closing but now it seems fine when i remove it. Maybe it was because I worked on some intermediate state or with some legacy package...
However, will remove it.

lib/statemanager.js
kumavis: fake-tx should be improved - it may have been broken as base tx was changed

lib/utils/receipt.js
kumavis: heh - consider some gentle linting

package.json
kumavis: at the very least you can target a specific commit
`git+https://github.com/nexusdev/provider-engine.git#commitHash`

kumavis: merged and published as `web3-provider-engine@8.1.15` 👍

lib/statemanager.js
mhhf: I made some ugly hacks for persisting transaction - mostly etherumjs native rlp serialisation is used for persistency. This means that no from value is saved. Instead the from value is inferred from the tx signature which is persisted.
But need was to be able to pretend the ownership of addresses to be able to explore hypothetic state together with persisting this state. Ethereumjs wasn't build with this need in mind, so we have either to modify ethereumjs (as I did ugly with this hacks) or come up with an own, clean solution. 
Persisting fake transactions was one of the major pain-in-the-ass tasks -.-

lib/statemanager.js
mhhf: This line is part of the ugly tx persistency hacks.
_getSenderAddress() generates the from value based on the tx signature. Because I had overwritten the default getSenderAddress functionality and no from value is given for raw transactions this is a necessary step to retrieve the senders address.

kumavis: sounds like we need to rewrite FakeTransaction
we can include the from address and drop the signature in the serialization format

lib/utils/txhelper.js
mhhf: I just followed the etheruemjs-tx style `_from` variable here which is faked in the same way. Again that is part of the ugly ethereumjs-tx hack.

package.json
mhhf: Thanks @kumavis, will include it here!
@tcoulter this was never intended to be part of the commit. But I needed a change for proper operation. Will change this!

package.json
mhhf: :+1:

test/accounts.js
mhhf: 👍  works without

lib/utils/forkedblockchain.js
mhhf: yes, this :)
I don't dig to deep into the relative vs. absolute block number and your code logic, just made enough changes to satisfy my need and make the tests pass. This part and the part above needs more consideration in oder for a clean refactoring.

lib/utils/forkedblockchain.js
mhhf: We should totally remove one. I was just sloppy on this.

lib/utils/forkedblockchain.js
mhhf: :+1:

lib/subproviders/reactiveblocktracker.js
mhhf: I had major trouble with it while working with logs on forked blockchain as some internal logic was triggering something twice. I don't remember fully what the side behavior was but after spending quite some time on debugging provider-engine (which is kind of hard) i just removed it here, and everything worked fine.

lib/statemanager.js
mhhf: Seems reasonable, will try to do this.

lib/statemanager.js
mhhf: I needed the rsv/ from information here for persistency which was not propagated after a signature check.

lib/statemanager.js
mhhf: On the second look I noticed a lot of problems with moving this logic into forkedBlockchain: The ForkedBlockchain/ BlockchainDouble classes don't have a getLogs function, only a getBlockLogs function. The main improvement with logForwarding in the forked range comes from batch logs: e.g. 0-latest. The only thing that I can think of is moving the whole getLogs function into ForkedBlockchain and BlockchainDouble which doesn't feels right.
Maybe its ok to let the StateManager be fork aware?!
Any suggestions on how to solve this, other then keeping it here?

lib/utils/forkedblockchain.js
```diff
@@ -73,8 +88,12 @@ ForkedBlockchain.prototype.initialize = function(accounts, callback) {
     self.web3.eth.getBlock(blockNumber, function(err, json) {
```
mhhf, line 88: I discovered a weirdness in web3 - if `typeof blockNumber === "string"` blockNumber is treaded as *hash* and a `eth_getBlockByHash` method call is generated which leads to a crash if block number 0 is requested which gets translated to `0x0`. Only in the case its an number, its treaded as the blockNumber.
I fixed it in my fork by introducing this line:
```
if(typeof blockNumber == "string" && blockNumber[1] === "x" && blockNumber.length < 42) blockNumber = parseInt(blockNumber)
```

However, it doesn't feel clean. Also there is a chance for this happening somewhere else - I haven't scanned the code for other occurrences of this (just noticed it here). Is this line _good enough_ or should I submit a new Issue for this?

package.json
mhhf: @tcoulter I updated the provider engine to the newest version (~8.1.16), However, this lead to the mining and mining_inverval tests failing.
The errors always happens at line like this: https://github.com/ethereumjs/testrpc/blob/master/test/mining.js#L300 
I suppose the old provider engine had the feature of throwing an error in case a throwing transaction was send, now its successfully getting included into the new block (I assume the user still pays for a failed gas). 

In my fork I skipped tests, which fails out of this reason.

lib/utils/forkedblockchain.js
mhhf: BlockchainDouble.createBlock() here is never called, instead BlockchainDouble builds on top of parentBlock, which we want to prevent here as we forking of with a new Block. super.createBlock() is needed for later in the ForkedBlockchain as well, so I cannot override this.
I think the best way is to keeping it here.

mhhf:
@tcoulter I submitted a bunch of new corrections, so you can go and have a look on them.
The biggest chunk I'm fighting with is ethereumjs-blockchain: Since my initial fork a lot of things were refactored, including breaking changes (such as having constrains on when save the block under its blockNumber, making querying arbitrary blocks by their number impossible) also I found the current state quite buggy (e.g. they forget to mark the blockchain as initialized after the initialization function finishes). I spend more then a full working day on trying to stabilize current ethereumjs-blockchain state and refactor it into testrpc and failed, with an unknown amount of work to go. But I've started talking about needed changes and bugs.
I'd suggest working with the current mod-blockchain fork until ethereumjs-blockchain is considered stable and all necessary features (such as flag for disable validator for forked blocks) are included.

mhhf:
Should I close this?

tcoulter:
Absolutely not! I really appreciate your work here, and I apologize for the length of time this has taken me to complete. Merging this is the TestRPC's highest priority, actually. We currently have people using it and noting memory issues which we need to explore. They may be unrelated to this change, but I need to sit down with it and check it out. I expect to do this in the next two weeks.

As an aside, I'm spread really thin. This has 100% been blocked by me and I take all the blame for it. I've sent you a message on gitter and would love to chat more.

tcoulter:
@mhhf Update:

I did some digging into this, and it looks like a) ethereumjs-blockchain adds restrictions that are hard to get around, and b) it's likely etheruemjs-blockchain is going to be rewritten. The general word is, "Ya, it isn't used that much." My impression of the conversation was that it wouldn't be a good idea to use it.

So instead, I used your branch as inspiration and rewrote database persistence in a different way, saving all the data in a custom fashion. You can see the PR [here](https://github.com/ethereumjs/testrpc/pull/275). 

The pros to this branch:

- The underlying data structures don't change. Data that were stored in arrays, like the `blocks` array, are still kept in array-like data structures. 

Cons: 

- All the data read/writes are now asynchronous, hence most of these changes. 

I'll probably be moving forward with the `persistence` branch within the next couple of days. All the tests pass except my poorly written performance test (I should have guessed it wouldn't pass). Feel free to put it through the ringer and let me know what you think.

tcoulter:
Closing this. Thanks for all your work @mhhf. Merged `persistence` into master.

Issue #190, opened 2016-10-26
eth_sign output is wrong when ecsign produces a signature whose 'r' starts with 0


cgewecke:
( @MrChico  Thank you for adding eth_sign to testrpc - it's super useful. )

It looks like leading zeros in the `r` and `s` buffer components are getting lost? 
## Expected Behavior

web3.eth.sign should always produce a hex prefixed string of length `132` 
## Current Behavior

Very occasionally - like 1 in 100 times with random string inputs  - eth_sign's output will have length `130` and running it through ethereumjs-util's `fromRpcSig` will produce an  undefined `v` value. 
## Possible Solution

Replace this block in `StateManager.prototype.sign`:

``` javascript
var r = utils.fromSigned(sgn.r);
var s = utils.fromSigned(sgn.s);
var v = utils.bufferToInt(sgn.v) - 27;
r = utils.toUnsigned(r).toString('hex');
s = utils.toUnsigned(s).toString('hex');
v = utils.stripHexPrefix(utils.intToHex(v));
return utils.addHexPrefix(r.concat(s, v));
```

with this statement:

``` javascript
return utils.toRpcSig(sgn.v, sgn.r, sgn.s);
```
## Steps to Reproduce (for bugs)

Start testrpc w/

```
--account="0xe6d66f02cd45a13982b99a5abf3deab1f67cf7be9fee62f0a072cb70896342e4"
```

Run

``` javascript
// Passes
var msg = "3c9229289a6125f7fdf1885a77bb12c37a8d3b4962d936f7e3084dece32a3ca1";
var msgHash = web3.sha3(msg);
var signed = web3.eth.sign( web3.eth.accounts[0], msgHash);
assert.equal(signed.length, 132 ) 

// Fails
var msg = "558d2681eeb61e8bd3ee590aa624a6739caf9bef529a3f6e63dc54459be3ebd1"
var msgHash = web3.sha3(msg);
var signed = web3.eth.sign( web3.eth.accounts[0], msgHash);
assert.equal(signed.length, 132 ) 
```
## Your Environment
- Version used: 2.2.7
- Node v.5.4.1
- Operating System and version: OSX El Capitan 10.11.5
- Link to your project: https://github.com/animist-io/whale-island

cgewecke:
Closed w/ 403e015ea7053e93ba36fe2b59c617b1a50f0d48

PR #191, opened 2016-10-27
Fix eth_sign bug (Issue #190)

cgewecke:
eth_sign currently fails if ecsign returns 'r' or 's' components that have leading zeros. This PR replaces existing logic that translates ecsign output to  rpc format with ethereumjs-utils `toRpcSig` method.


tcoulter:
Nice job, thanks!

Issue #197, opened 2016-11-02
ethereumjs-vm 2.0.1 compatibility.


simondlr:
I upgraded to use the latest version of the vm (2.0.1). However, with it came issues wrt state trie lookups. Getting an issue wrt funds not being available.

```Error: sender doesn't have enough funds to send tx. The upfront cost is: 471238800000000000 and the senders account only has: 0```

Thanks to @mhhf (Denis), he pointed me in the right direction. It's a simple fix since the interface of the VM changed.

At blockchain_double.js, change the following part from:

```
  this.vm = options.vm || new VM(this.stateTrie, this, {
    enableHomestead: true,
    activatePrecompiles: true
  });
```

to

```
  this.vm = options.vm || new VM({
    state: this.stateTrie,
    enableHomestead: true,
    activatePrecompiles: true
  });
```

This correctly passes the correct state trie.

I can submit a PR, but it seems there are more extensive code/writing around blockchain state in testrpc, so not sure what takes precedence atm. Let me know.

**Comment by simondlr** [comment]
I added a PR anyway. :)

**Comment by tcoulter** [comment]
Most tests pass with your PR. I need to research the three failing tests; they all revolve around a specific blockhash that's generate. It could just be those tests were poorly written.

**Comment by jmdfm** [comment]
Here are the 3 failing tests in question: 

```
150 passing (13s)
  3 failing

  1) Forking should represent the block number correctly in the Oracle contract (oracle.blockhash0), providing forked block hash and number:

      Uncaught AssertionError: '0x044852b2a670ade5407e78fb2863c51de9fcb96542a07186fe3aeda6bb8a116d' == '0x1e27de32f8daf6d4adc903b68282e866dd6bdf3409d4b4976885f4cfa5dd399e'
      + expected - actual

      -0x044852b2a670ade5407e78fb2863c51de9fcb96542a07186fe3aeda6bb8a116d
      +0x1e27de32f8daf6d4adc903b68282e866dd6bdf3409d4b4976885f4cfa5dd399e

      at test/forking.js:373:18

  2) Provider: contract scenario should represent the block number correctly in the Oracle contract (oracle.blockhash0):

      Uncaught AssertionError: '0x044852b2a670ade5407e78fb2863c51de9fcb96542a07186fe3aeda6bb8a116d' == '0x6dd2f0ee6e56750dfb0bc20f9dab7a1833def6c5dd17479d495b3fe95d84ac2c'
      + expected - actual

[...]

PR #198, opened 2016-11-02
Upgrade to ethereumjs-vm 2.0.1

simondlr:
See: #197.

### DIFF
--- a/lib/blockchain_double.js
+++ b/lib/blockchain_double.js
@@ -21,7 +21,8 @@ function BlockchainDouble(options) {
 
   this.stateTrie = options.trie || new Trie();
 
-  this.vm = options.vm || new VM(this.stateTrie, this, {
+  this.vm = options.vm || new VM({
+    state: this.stateTrie,
     enableHomestead: true,
     activatePrecompiles: true
   });
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "ethereumjs-block": "^1.2.2",
     "ethereumjs-tx": "^1.1.0",
     "ethereumjs-util": "^4.5.0",
-    "ethereumjs-vm": "^1.3.0",
+    "ethereumjs-vm": "^2.0.1",
     "ethereumjs-wallet": "^0.6.0",
     "fake-merkle-patricia-tree": "^1.0.1",
     "heap": "^0.2.6",


redsquirrel:
This resolves a problem that has been plaguing a project that @drcode have been working on. It manifested itself as an "out of gas" error. I just confirmed that this PR resolves our issue! 🎉

tcoulter:
Is this still in progress? (it's tagged as "in progress"). Will pull and run the tests.

simondlr:
Oh. Sorry. Thought "in progress" meant, as in, reviewing "in progress".

tcoulter:
:+1: I don't know what it means. :) I thought it meant you're still working on it.

Either way, upon first look it appears ethereumjs-vm might have an issue when pulling the blockhash within solidity code. Need to take a deep dive before this can be merged.

simondlr:
> I thought it meant you're still working on it.

Oh, it was added automatically. Might be github doing this. :D

> Either way, upon first look it appears ethereumjs-vm might have an issue when pulling the blockhash within solidity code.

Yeah, might be related to some other changes.

tcoulter:
@johnmcdowall I've assigned this one to you. I've created a new branch called `simondlr-master` that includes Simon's work and a couple test fixes. 

There are still some tests that fail which need to be looked at before we can merge this branch. From what I can tell, the tests that fail are all related to processing contracts that asks for a block's hash from within the contract code itself. Somehow the block hash returned in code doesn't match what we expect it to be. My hunch is this is a problem within the newest ethereumjs-vm itself; if so, it should lead you to the `BLOCKHASH` opcode here: https://github.com/ethereumjs/ethereumjs-vm/blob/master/lib/opFns.js#L326. When I've previously debugged this, I've used `npm link` on my own version of ethereumjs-vm to better understand the code the `BLOCKHASH` opcode is calling (and hence why the wrong hash is returned). That said, this could totally be an issue with the TestRPC as well.

The main work here is a bunch of debugging. Getting a handle on all the moving parts is likely the hardest. Let me know if you have any questions.

tcoulter:
@johnmcdowall PS: [This](https://github.com/ethereumjs/testrpc/blob/master/test/requests.js#L392) is the test you should fix first. There's an identical test that performs the same actions but within the context of chain forking; ignore that one. Fixing the non-forked version of the test should likely fix the forked one as well.

jmdfm:
Right, so the new interface as well as changing to opts, made passing a blockchain an option where before it was required. This is why the 3 tests are failing, we're not passing in the blockchain double as the `blockchain` opt, so the VM is creating one every test run which results in block 0 having a different hash every time.

jmdfm:
See PR #205

lib/blockchain_double.js
```diff
@@ -21,7 +21,8 @@ function BlockchainDouble(options) {
 
   this.stateTrie = options.trie || new Trie();
 
-  this.vm = options.vm || new VM(this.stateTrie, this, {
+  this.vm = options.vm || new VM({
+    state: this.stateTrie,
```
axic, line 25: The `this` parameter was missed, it should be here as `blockchain: this,`.

jmdfm: Haha you just jinxed me on PR #205 :)

jmdfm:
Closing this in favour of #205 which is ready

PR #201, opened 2016-11-10
Be stricter with the package.json version matching (uses ~ to match only recent minor versions)

jmdfm:


jmdfm:
@tcoulter Merge at will Tim

tcoulter:
:boom:

PR #211, opened 2016-11-29
pad position left to 32 bytes before looking up in trie

aakilfernandes:
Fixes https://github.com/ethereumjs/testrpc/issues/133


tcoulter:
Thanks @aakilfernandes!

PR #212, opened 2016-11-29
include_full_transactions fix

aakilfernandes:
`getBlockByHash` and `getBlockByNumber` should only return full transaction objects when the 2nd parameter is true. web3 sets it to false by default.


tcoulter:
Nice. Thanks @aakilfernandes.

PR #218, opened 2016-12-05
adding some verbose logging and error logging

cicorias:
This adds some logging when verbose is on (config.verbose) and also in an error situation when the request is received, it uses the logger.error method that was added to the code; which only had logger.log prior. 

This has been helpful for debugging client calls in some things we are testing.

### DIFF
--- a/lib/interface.js
+++ b/lib/interface.js
@@ -21,7 +21,8 @@ var Interface = {
 
     if (options.logger == null) {
       options.logger = {
-        log: function() {}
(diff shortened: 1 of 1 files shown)


benjamincburns:
Sorry, but the logic you're patching has moved to [trufflesuite/ganache-core](https://github.com/trufflesuite/ganache-core). If you still feel this change is necessary, can you please reopen this PR against the develop branch in that repo?

PR #223, opened 2016-12-14
test balance of new contract

aakilfernandes:
All this does is add a test for deploying contracts with a value. Had an issue locally but looks like its not TestRPC related.


tcoulter:
I'm not sure why but this change adds significant test failures. We need to look into these before merging.

tcoulter:
The contract's constructor needed to be `payable` in order to send value. Fixed. Thanks!

PR #229, opened 2016-12-28
change gasPrice default to 20000000000 in readme

aakilfernandes:
Based on this https://github.com/ethereumjs/testrpc/blob/eb27c909077db454901af2e53e3a631d3973a575/test/requests.js#L105-L114

Also 20000000000 when testing locally


tcoulter:
Good catch!

PR #231, opened 2016-12-30
Test nonce too low

aakilfernandes:
Wrote this test in the hunt of another bug. Doesn't actually change anything, just adds an additional test.


tcoulter:
Thanks @aakilfernandes!

PR #232, opened 2016-12-30
sendRawTransaction respect 0 buffers and include nonce

aakilfernandes:
A couple of issues this PR fixes

1) including the nonce when forming raw transactions
https://github.com/ethereumjs/testrpc/compare/rawtx-nonce-issue?expand=1#diff-1c54b13873a64b628fdff8921872d262R222

2) Getting the raw transaction pipline to respect empty buffers (`<Buffer >`). This is because `<Buffer >` is falsy, even though it means `0`. For example, if I use a nonce of 0 or gas of 0, pass that along instead of overriding with a `null`

### DIFF
--- a/lib/statemanager.js
+++ b/lib/statemanager.js
@@ -213,12 +213,13 @@ StateManager.prototype.queueRawTransaction = function(rawTx, callback) {
 
   var tx = new FakeTransaction(data);
   var txParams = {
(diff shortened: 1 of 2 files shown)


aakilfernandes:
Actually, I'm going to submit a slightly simplified version of this pr. closing for now

PR #233, opened 2016-12-30
fix sendRawTransaction nonce and gasLimit issues

aakilfernandes:
1) Check .gasLimit instead of .gas
https://github.com/ethereumjs/testrpc/compare/rawtx-nonce-issue-simplified?expand=1#diff-1c54b13873a64b628fdff8921872d262R218

2) Include nonce
https://github.com/ethereumjs/testrpc/compare/rawtx-nonce-issue-simplified?expand=1#diff-1c54b13873a64b628fdff8921872d262R222

3) Respect zero buffer `<Buffer >` for raw transactions instead of inducing to `null`. This is to allow 0 nonces/gasPrice/gasLimit.

### DIFF
--- a/lib/statemanager.js
+++ b/lib/statemanager.js
@@ -213,12 +213,13 @@ StateManager.prototype.queueRawTransaction = function(rawTx, callback) {
 
   var tx = new FakeTransaction(data);
   var txParams = {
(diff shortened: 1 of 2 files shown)


tcoulter:
Awesome, thanks @aakilfernandes. Easy to review when all tests pass. Cheers!

PR #240, opened 2017-01-12
Bundling of testrpc server and ethereumjs-testrpc

juanfranblanco:
Node bundling to simplify "installation" on different OSs.

* Bundling browserifies excluding builtins (no shims)
* Builds a module
* includes __dirname and __filename and proccess

Portable server bundling to simplify running a server on different OSs (just download testrpc-portable.js and run it with node)

* Bundling browserifies excluding builtins (no shims)
* includes __dirname and __filename and proccess

Browserify defaults web3-provider-engine/subproviders/solc.js as false using the project.json setting
this allows to be ignored if set to blank. This is gracefully handled on the code by catching an exception when is blank


lib/interface.js
```diff
@@ -154,7 +154,11 @@ var Interface = {
     engine.addProvider(new FilterSubprovider());
     engine.addProvider(new GethDefaults());
     engine.addProvider(new VmSubprovider());
-    engine.addProvider(new SolcSubprovider());
+    try {
+    engine.addProvider(new SolcSubprovider());  
+    } catch (error) {
+      //graceful failover for browser, has to be a better way. 
+    }
```
tcoulter, line 161: Interesting. There's no way to browserify solc?

package.json
```diff
@@ -25,6 +25,7 @@
     "fake-merkle-patricia-tree": "~1.0.1",
     "heap": "~0.2.6",
     "merkle-patricia-tree": "~2.1.2",
+    "scryptsy": "^2.0.0",
```
tcoulter, line 28: Can you tell me what this import is used for? It doesn't look obviously used within this PR.

tcoulter (commented):
Very cool. Added a couple questions for you.

package.json
juanfranblanco: Totally my fault, I left that after the original experimentation. Not needed.

lib/interface.js
juanfranblanco: I have not found any way. @SilentCicero created originally an issue with christian a couple of months ago and did not get very far. We could investigate further onto this.

juanfranblanco: But it works in browser solidity (confusing)

fanatid:
This files need to be updated after every new version, maybe there some way to automatize this? For example with CI service (travis) which will build and upload new version to specific branch?

benjamincburns:
Closing this PR, as I think it no longer applies? Please sing out if this is a mistake!

Issue #243, opened 2017-01-19
eth_sign API out of sync with ethapi


parkan:
As of https://github.com/ethereum/go-ethereum/pull/2940, eth_sign api has changed. It is now `eth_sign(address, msg)`, accepting arbitrary input and putting it through keccak256.

 

## Expected Behavior

Upstream eth_api introduced a breaking change where `eth_sign` now accepts the actual *message*, rather than a 32 byte hash digest. Signature is now `eth_sign(address, msg)`

## Current Behavior

testrpc currently emulates existing `eth_sign(address, sha3(msg))` behavior

## Possible Solution

Change `ethereumjs-util.ecsign` to call `keccak256` before calling `secp256k1.sign`

## Context

We need to sign statements with web3, which is now incompatible between testrpc and real ethapi.

## Your Environment

* Version used: v3.0.3
* Environment name and version (e.g. PHP 5.4 on nginx 1.9.1): n/a
* Server type and version:
* Operating System and version: os x
* Link to your project: https://github.com/mediachain/aleph-ethereum

yusefnapora:
I believe this may also need to prefix `msg` with `Buffer.from('\u0019Ethereum Signed Message:\n' + msg.length.toString())` before running through `keccak256`, to be compatible with the eth_api implementation

parkan:
@yusefnapora good point, I mostly copied over the impl notes from the Parity issue, which omitted that

fumecow:
I made a diagram of the current differences between testrpc, geth and parity with respect to eth_sign behavior and on-chain ecrecover.

![signature generation validation](https://cloud.githubusercontent.com/assets/27297969/25290626/14f00a96-268c-11e7-925f-6c5c1b4be485.png)

tcoulter:
Wow, that's a great diagram. If you have it handy, could you give me some messages, a test private key / mnemonic and the signed results of geth/parity for testing?

fumecow:
Unfortunately accounts differ when testing on the different nodes so an apples to apples comparison isn't really possible.  I have some test script output that gives some data but missing is the private key for the account:

Using node=EthereumJS TestRPC
Using account: 0x264821d83625ed0cb177bb01a486e304cdece092
Using contract: 0x59e4afee0cc669cfbcefd894429e27d2a839c45d
Generating signature
  address=0x264821d83625ed0cb177bb01a486e304cdece092
[...]

PR #244, opened 2017-01-19
Make web3.eth.sign compatible with new eth_sign API

yusefnapora:
As discussed in #243, the eth_sign api now accepts an arbitrary message, which is prefixed with a fixed header + the length of the message, then run through keccak256 before signing.

This updates the `sign` method in `lib/statemanager.js` to produce the new signatures, and updates the tests in `test/requests.js` - I managed to find another message that produces a signature with a zero-prefixed `r` component, so hopefully that test is still valid.

Actually recovering the public key / address for a signature is quite a bit more involved now, but that's due to the API change.  Price of progress, I guess.

Let me know if anything needs tweaking.


axic:
I think it would be great adding this functionality to the underlying library, `ethereumjs-util`. Would you be interested looking into that?

yusefnapora:
I opened the PR in ethereumjs-util - I'm happy to update this one to use it once it gets merged.

yusefnapora:
Now that there's a new version of ethereumjs-util with the `hashPersonalMessage` helper, I updated this PR to use it.  

I had to make a couple small changes to get the tests passing again with ethereumjs-util 5.x - the API for `ecrecover` changed, and I guess maybe `bufferToInt` got more strict, because it was throwing during `txhelper.fromJSON` complaining that the input was > 53 bits.

Let me know if this needs anything else :)

parkan (approved):
This API divergence is breaking our build, any chance this can get merged?

tcoulter:
Apologies for the delay. Thanks a lot @yusefnapora!

yusefnapora:
Sweet, glad to help 😄

PR #251, opened 2017-02-06
Add networkId option to README options

izqui:


axic:
This seems to be duplicating #256 Can you discuss with @logvinovleon the final wording?

tcoulter:
Added in a different PR. Thanks!

Issue #252, opened 2017-02-06
transactionIndex is always 0


mpolci:
I started testrpc with the option -b 10 to mine a block every 10 senconds.
If I have many transactions in the same block, all transactions have the transactionIndex field equal to 0.

mpolci:
see PR https://github.com/ethereumjs/testrpc/pull/253

benjamincburns:
This should have been closed when Tim merged the PR. Thanks again for your contribution, @mpolci!

PR #253, opened 2017-02-06
transactionIndex field on transaction and transactionReceipt. Resolve issue #252

mpolci:


lib/utils/receipt.js
```diff
@@ -11,8 +11,8 @@ function Receipt(tx, block, logs, gasUsed, contractAddress) {
 
   for (var i = 0; i < block.transactions.length; i++) {
     var current = block.transactions[i];
-    if (current.hash() == tx.hash()) {
```
axic, line 14: Would using `===` be enough?

tcoulter:
Thanks!

lib/utils/receipt.js
mpolci: The `hash()` function returns an Uint8Array and you cannot use the equality operator to compare arrays, you have to use the `equals()` member function

PR #259, opened 2017-02-14
Adds primitive support for WebSockets.

### LINKED ISSUE #257: Support WebSocket as a transport for the RPC.
Geth and Parity both support WebSockets (to some extent) for the transport layer for the Ethereum RPC.  It would be nice if TestRPC also supported WebSockets so one can test using it when writing an app that communicates via WebSockets.

MicahZoltu:
This won't work for pub/sub stuff, but it at least gets things working for request/response over WebSockets.

Fixes #257

### DIFF
--- a/lib/interface.js
+++ b/lib/interface.js
@@ -1,4 +1,5 @@
 var http = require("http");
+var WebSocket = require('ws');
 
 var ProviderEngine = require("web3-provider-engine");
 var FilterSubprovider = require('web3-provider-engine/subproviders/filters.js');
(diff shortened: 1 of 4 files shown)


lib/interface.js
```diff
@@ -128,6 +129,50 @@ var Interface = {
...
+    var logger = options.logger;
+    var provider = this.provider(options);
+    var webSocketServer = new WebSocket.Server(options.webSocketServer);
+    webSocketServer.on('connection', (webSocket) => {
+      webSocket.on('message', (messageJson) => {
+        var message = JSON.parse(messageJson);
+        switch (message.method) {
+          case "eth_subscribe":
+            // TODO: add support for pub-sub, at the moment this just returns a token that will never get any notifications
+            webSocket.send(JSON.stringify({ jsonrpc: "2.0", id: message.id, result: "0xcd0c3e8af590364c09d0fa6a1210faf5" }));
```
MicahZoltu, line 160: I'm not sure what the right way to do stubbed pub-sub is, but for now this satisfies the protocol if someone tries to subscribe.

package.json
```diff
@@ -30,13 +30,15 @@
...
+    "ws": "2.0.3",
     "yargs": "~3.29.0"
   },
   "devDependencies": {
     "eslint": "~3.0.1",
     "eslint-config-standard": "~5.3.5",
     "eslint-plugin-standard": "~1.3.3",
-    "mocha": "~2.2.5"
+    "mocha": "~2.2.5",
+    "@parity/parity.js": "0.3.81"
```
MicahZoltu, line 41: This is required because `web3` doesn't support web sockets yet and I needed a "Ethereum standards based" library to test that the transport is correctly conforming to what is expected.  I'm open to ideas on other libraries if `parity.js` isn't desired.

test/websocket.js
```diff
@@ -0,0 +1,46 @@
...
+  // parity.js is designed for browsers where `WebSocket` is globally accessible
+  var originalGlobalWebSocket;
+  before(() => {
+    originalGlobalWebSocket = global.WebSocket;
+    global.WebSocket = require('ws');
+  });
+  after(() => {
+    global.WebSocket = originalGlobalWebSocket;
+  });
+  // END terribleness
```
MicahZoltu, line 17: Ugh.  `parity.js` is designed for usage in browser, where there is a global `WebSocket` class available.  Since this test is running in node, we have to make `global.WebSocket` available for the duration of these tests.  The `after` is there to ensure this test's behavior doesn't leak into other tests.

test/websocket.js
```diff
@@ -0,0 +1,46 @@
...
+    return api.net.version().then((version) => {
+      assert.strictEqual(version, "5");
+    });
+  });
+
+  it("eth.gasPrice", () => {
+    return api.eth.gasPrice().then((gasPrice) => {
+      assert.deepEqual(gasPrice, { s: 1, e: 10, c: [ 20000000000 ] });
+    });
+  });
```
MicahZoltu, line 45: I can add more tests beyond just `net.version` and `eth.gasPrice` if desired.  I mainly wanted to just make sure that a couple simple things worked and could run in sequential tests (adding the second test found a number of bugs).  I'm not sure what exactly we would want to test without going through and testing the entire RPC from within this spec...

MicahZoltu:
Fixes #257

benjamincburns:
Closing this, as this logic has moved to [trufflesuite/ganache-core](https://github.com/trufflesuite/ganache-core), and there's a similar PR for this over there (trufflesuite/ganache-core#14).

PR #261, opened 2017-02-14
Initial implementation of unix socket IPC support [WIP]

### LINKED ISSUE #180: Support IPC endpoint
Some tools (i.e. Mist, Solidity tests, etc.) have switched to use IPC instead of the HTTP-RPC.

Would it be hard adding IPC support? (That would need to replace the http server when the option is turned on.)

axic:
Fixes #180.

### DIFF
--- a/lib/interface.js
+++ b/lib/interface.js
@@ -1,4 +1,6 @@
 var http = require("http");
+var net = require("net");
+var fs = require("fs");
 
 var ProviderEngine = require("web3-provider-engine");
 var FilterSubprovider = require('web3-provider-engine/subproviders/filters.js');
@@ -126,6 +128,98 @@ var Interface = {
     server.provider = provider;
 
     return server;
+  },
+
+  ipcServer: function(options) {
+    if (options == null) {
+      options = {};
+    }
+
+    if (options.logger == null) {
+      options.logger = {
+        log: function() {}
+      };
+    }
+
+    var logger = options.logger;
+    var provider = this.provider(options);
+
+    var server = net.createServer(function(socket) {
+      socket.on('close', function () {
+        console.log('close');
+      });
+      socket.on('connect', function () {
+        console.log('connect');
+      });
+      socket.on('data', function(data) {
+        console.log("data", data.toString());
+
+        var payload;
+        try {
+          payload = JSON.parse(data);
+        } catch(e) {
+          // NOTE: some clients expect the trailing new line
+          socket.write('{"jsonrpc": "2.0", "id": null, "error": {"code": -32700, "message": "Bad request"}}\n');
+          return;
+        }
+
+        // Log messages that come into the TestRPC
+        if (payload instanceof Array) {
+          // Batch request
+          for (var i = 0; i < payload.length; i++) {
+            var item = payload[i];
+            logger.log(item.method);
+          }
+        } else {
+          logger.log(payload.method);
+        }
+
+        provider.sendAsync(payload, function(err, result) {
+          // NOTE: some clients expect the trailing new line
+          socket.write(JSON.stringify(result) + '\n');
+        });
+      });
+      socket.on('drain', function() {
+        // emitted when the write buffer becomes empty
+        console.log('drain');
+      });
+      socket.on('end', function() {
+        console.log('end');
+      });
+      socket.on('error', function () {
+        console.log('error');
+      });
+      socket.on('lookup', function () {
+        console.log('lookup');
+      });
+      socket.on('timeout', function () {
+        console.log('timeout');
+      });
+    });
+
+    server.on('close', function () {
+      console.log('server close');
+    });
+    server.on('connection', function () {
+      console.log('server connection');
+    });
+    server.on('error', function () {
+      console.log('server error');
+    });
+    server.on('listening', function () {
+      console.log('server listening');
+    });
+
+    fs.unlink(options.ipcPath);
+    return {
+      listen: function (port, hostname, callback) {
+        server.listen(options.ipcPath);
+
+        provider.manager.waitForInitialization(function(err, accounts) {
+          callback(err, accounts);
+        });
+      }
+    }
   },
 
   provider: function(options) {
--- a/README.md
+++ b/README.md
@@ -47,6 +47,8 @@ Options:
 * `-l` or `--gasLimit`: Use a custom Gas Limit (defaults to 0x47E7C4)
 * `-f` or `--fork`: Fork from another currently running Ethereum client at a given block. Input should be the HTTP location and port of the other client, e.g. `http://localhost:8545`. You can optionally specify the block to fork from using an `@` sign: `http://localhost:8545@1599200`.
 * `--debug`: Output VM opcodes for debugging
+* `--ipc`: Run in IPC mode
+* `--ipcPath`: Set the IPC path (defaults to ethereum.ipc)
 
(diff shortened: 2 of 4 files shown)


axic:
Very rudimentary at the moment, but works.

lib/interface.js
```diff
@@ -128,6 +130,97 @@ var Interface = {
...
+
+        var payload;
+        try {
+          payload = JSON.parse(data);
+        } catch(e) {
+          socket.write('{"error":"Bad request"}');
+          return;
+        }
+
+        // Log messages that come into the TestRPC via http
```
MicahZoltu: Comment should probably say `via IPC` rather than `via http`.  Copy paste issue I suspect?

lib/interface.js
```diff
@@ -128,6 +130,97 @@ var Interface = {
...
+        console.log('connect');
+      });
+      socket.on('data', function(data) {
+        console.log("data", data.toString());
+
+        var payload;
+        try {
+          payload = JSON.parse(data);
+        } catch(e) {
+          socket.write('{"error":"Bad request"}');
```
MicahZoltu: Down on line 181 there is a comment suggesting that some clients expect a trailing new line.  Should that be applied here as well?

MicahZoltu: Consider including the error in the response so the user can have an easier time understanding what their problem was:
```
socket.write('{"error":"Bad request","message":"' + e + '"}\n')
```

MicahZoltu (changes requested):
Really the only thing blocking a merge (given my terribly limited understanding of this project or IPCs in general) is the missing newline on the failure response, and possibly inclusion of the parsing error in the response.

MicahZoltu (approved):

lib/interface.js
```diff
@@ -128,6 +130,98 @@ var Interface = {
...
+    });
+    server.on('error', function () {
+      console.log('server error');
+    });
+    server.on('listening', function () {
+      console.log('server listening');
+    });
+
+    fs.unlink(options.ipcPath);
+    return {
```
MicahZoltu, line 214: Recommend making this return the underlying server.  This is how the HTTP TestRPC server is implemented and I believe consistency is important.

MicahZoltu (changes requested):

lib/interface.js
MicahZoltu: In particular, the way you have done it here means someone cant' call `.close()` on the returned object.  You could just add a passthrough close method, but that doesn't feel like the right solution.

MicahZoltu: My thought is something like this to maintain consistency with the HTTP server implementation.  I'm not a fan of this style of monkeypatching functions to change from synchronous to asynch, but consistency is important here.
```
  var underlyingServerListen = server.listen;
  server.listen = function (ipcPath, callback) {
    underlyingServerListen(ipcPath);
    server.provider.manager.waitForInitialization(function(error, accounts) {
      callback(error, accounts);
    });
  };

  return server;
```

MicahZoltu: Hmmm, this pattern doesn't match up well with Web Sockets as they don't have a "listen" step.  Constructing the WebSocket server binds to the port immediately and the options must be passed into the constructor.  If we accept that the pattern won't fit for all 3 transports, is it worth bothering with consistency?  Perhaps all 3 should bind to their respective ports on construction and the "listen" method can be changed to "start" or something and would just initialize the backing provider.

lib/interface.js
```diff
@@ -128,6 +130,98 @@ var Interface = {
...
+    server.on('connection', function () {
+      console.log('server connection');
+    });
+    server.on('error', function () {
+      console.log('server error');
+    });
+    server.on('listening', function () {
+      console.log('server listening');
+    });
+
```
MicahZoltu, line 212: For consistency, `server.provider` should be set to `provider`, like the HTTP one.

lib/interface.js
```diff
@@ -128,6 +130,98 @@ var Interface = {
...
+      });
+      socket.on('connect', function () {
+        console.log('connect');
+      });
+      socket.on('data', function(data) {
+        console.log("data", data.toString());
+
+        var payload;
+        try {
+          payload = JSON.parse(data);
```
MicahZoltu, line 159: Sadly, I don't believe this will work.  From what I can tell, Ethereum JSON-RPC over IPC doesn't define any message boundary or enveloping.  JSON-RPC (the general one) assumes that the transport defines message boundaries but unlike HTTP and WS, IPC is just a byte stream so it is up to the protocol to define message boundaries.  I believe this is why the newline is required by some clients, though I have to wonder if those clients will do-the-right-thing if a message has newlines in it.

Best option I can come up with to correctly handle this is to have a streaming JSON parser that streams until it receives a complete single object.  Once it does, it would call that a single message and then start streaming until it has received another object.

I'm open to better ideas, I am doing some related work and would love a better solution.  :/

tcoulter:
@MicahZoltu Hey, I'm finally getting back to the TestRPC (and the Truffle team has more staff now, thank goodness). Can I get you to evaluate your above comments and see if they hold true? I don't know enough about IPC either to try this out, but I will put more time into it if you believe the above is still not coded correctly. Appreciate it!

MicahZoltu:
I believe the comments still apply, though at this point my _real_ recommendation is to not support IPC at all and instead only support WS and HTTP.

I think the inclusion of IPC support, which pushes message enveloping to the protocol, in Geth and Parity was a mistake.  With a JavaScript based node any performance benefits of IPC over WS are going to be so small in comparison to the fact that you are running your node in JavaScript that there is really no compelling reason to use IPC.  The one caveat would be if you want to use TestRPC to test your dApp's IPC code.  However, I think that in general people should stop using IPC in their dApps unless they have a _really_ compelling reason to do so (like you need those extra nanoseconds).  I haven't seen any particularly compelling arguments for IPC over WS at this point, which is why I would push to discourage its usage across the ecosystem.

benjamincburns:
Closing this as the logic has moved to [trufflesuite/ganache-core](https://github.com/trufflesuite/ganache-core) now. If this is still relevant, please reopen as a PR against the `develop` branch there.

Also, @tcoulter, we should likely discuss @MicahZoltu's last comment above.

axic:
It is very much relevant given every tool has moved on to IPC and are going to (or have) shut down HTTP support. Though I do not have any resources to work on this.

MicahZoltu:
Tools are preferencing WS over HTTP, which I can get behind.  I'm unaware of any tools moving _towards_ IPC though.  Do you have some examples?  I know Geth and Parity initially implemented IPC (and due to the lack of spec implemented it differently) but both now support WS which is "fast enough" for just about every use case I have been able to come up with and it provides message boundaries (unlike IPC) which means authoring tooling against it is significantly simpler (no need for a streaming JSON parser that can handle a stream of non-delimited JSON objects).

axic:
Should have written IPC + WebSockets, with heavier weight on WebSockets. Either way, regular HTTP is not really preferred going forward.

benjamincburns:
@axic no worries if you don't have the resources - it's on my list.

benjamincburns:
Provided this still merges cleanly we should probably just pull the `ipc` branch into `ganache-core` and work on it from there

adipascu:
This would be a cool feature to have, I am interested in having Ganache as a drop-in replacement for my local geth/parity setup.

PR #264, opened 2017-02-15
Implement cpp-ethereum's test RPC API [WIP]

axic:
This API is used by Solidity's soltest (the testing architecture) and https://github.com/ethereum/ethereum-console

Also see: https://github.com/ethereum/interfaces/issues/4

### DIFF
--- a/lib/subproviders/geth_api_double.js
+++ b/lib/subproviders/geth_api_double.js
@@ -298,7 +298,7 @@ GethApiDouble.prototype.miner_stop = function(callback) {
 
 GethApiDouble.prototype.rpc_modules = function(callback) {
   // returns the availible api modules and versions
-  callback(null, {"eth":"1.0","net":"1.0","rpc":"1.0","web3":"1.0","evm":"1.0"});
+  callback(null, {"eth":"1.0","net":"1.0","rpc":"1.0","web3":"1.0","evm":"1.0","test":"1.0"});
 };
 
 /* Functions for testing purposes only. */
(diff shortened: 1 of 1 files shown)


lib/subproviders/geth_api_double.js
```diff
@@ -322,6 +322,62 @@ GethApiDouble.prototype.evm_mine = function(callback) {
   });
 };
 
+/* cpp-ethereum specific methods, used by Solidity tests */
 
+var test_blocks = {}
```
axic, line 327: Basically this API has the ability to roll back to any previous block. That would mean to keep a snapshot of every block during the generic mine method.

lib/subproviders/geth_api_double.js
```diff
@@ -322,6 +322,62 @@ GethApiDouble.prototype.evm_mine = function(callback) {
...
+
+GethApiDouble.prototype.test_setChainParams = function(chain, callback) {
+  var self = this;
+  Object.keys(chain.accounts).forEach(function(address) {
+    var account = self.state.accounts[address];
+    if (account) {
+      // FIME: support changing balance of associated account
```
axic, line 334: This needs to be implemented.

tinybike:
@axic I'm curious, is cpp-ethereum still under active development / is it a client that is reasonably widely used?  All I hear about these days is geth and parity...

axic:
@tinybike it is still actively maintained, but I don't think it is widely used

axic:
This API is used by Solidity's soltest (the testing architecture) and https://github.com/ethereum/ethereum-console

benjamincburns:
@axic how do you rate this suite against something like [karalabe/hive](https://github.com/karalabe/hive/)? Would it be worth the effort to move this sucker over to [trufflesuite/ganache-core](https://github.com/trufflesuite/ganache-core), or would it be better if we concentrated our efforts on `hive`? I'm somewhat inclined to think the latter.

Closing either way.

axic:
This has no relation to hive, it has a different purpose.

benjamincburns:
@axic sorry, I misunderstood the basis for this PR... Rereading, the real question I should've asked is how relevant this still is today? Do the solidity unit tests still rely on these RPC calls in cpp-ethereum?

@tcoulter said that you guys are testing against testrpc already today (since we broke the solidity tests last month)? How is this accomplished? Does this change make that task easier?

axic:
> Rereading, the real question I should've asked is how relevant this still is today? Do the solidity unit tests still rely on these RPC calls in cpp-ethereum?

Yes.

> said that you guys are testing against testrpc already today

No. We are running Zeppelin's tests, which uses truffle internally. But that only runs Zeppelin's tests cases and not Solidity cases (~ 600 hundred tests).

> Does this change make that task easier?

Has no relation to Zeppelin.

PR #270, opened 2017-03-03
Stub for eth_protocolVersion

nanexcool:
Implements RPC method `eth_protocolVersion` returning `-1`.

This is just so we don't get an error as some of our internal tools call this method.


tcoulter:
This has been added to the `release` branch of trufflesuite/ganache-core. Will be added to the next release of the TestRPC. Thanks!

PR #275, opened 2017-03-09
Persistence

tcoulter:
Add persistence to the TestRPC. Uses levelup and a custom db structure for storing data. Some important items:

- `./lib/database` is where the data structure is defined.
- Data is serialized to JSON *except* for data stored directly by the state trie.
- I've written serializers for specific data types, as well as adapters to make levelup key/value pairs feel like arrays. Props to the `level-sublevel` module for making this a breeze. 
- Data is no longer stored in a big array in memory! Instead, it's **always on disk**. When you don't specify a db path, one is provided for you in a temp directory. This adds a cut to performance, but what's lost in speed is gained in stability. This **should** cut down on the memory issues (since nothing is stored in memory for long periods), and there was an [odd error](https://github.com/ethereumjs/testrpc/issues/81) that kept cropping up that db persistence seems to remove completely.
- We should be able to browserify the database the filesystem for use in the browser. Will be looking into that soon.
- I've broken up the provider and the server into their own separate files.
- Added a persistence tests, a quick performance test and tests for snapshotting.


tcoulter:
Looks like the tests fail because my perf test is highly dependent on the environment its running in. Go figure.

tcoulter:
:tada: Confirmed memory issues are fixed!

noeleon930:
Hi @tcoulter, this feature is so useful, thanks a lot.

But I think just found a tiny problem in this feature, when I assign a db_path like this:
```--db="~/testrpc-db"``` 

It seems like levelup can't understand what `~` is, and any relative path,
but  assigning an absolute path is working.

Environment:
        node -v = 6.10.0
        OS = Windows 10, macOS, Ubuntu 16.04
        branch = master (@03a401adc4a6dc1ea19679b9acf025d8231028aa)

Thanks!

PR #283, opened 2017-03-28
stop_server testrpc function to allow integration with test suites like karma

denisgranha:
In nodejs we can import the testrpc provider as a library, but in javascript this cannot be done, it needs to run a server instance, so this function allows to use testrpc integrated with karma and start/stop the server on each test.


benjamincburns:
@denisgranha - if you still need this functionality, can you please reraise this PR against the `develop` branch in [trufflesuite/ganache-core](https://github.com/trufflesuite/ganache-core)?

Also can you please submit a test along with the change?

PR #291, opened 2017-04-10
Add header to cache CORS requests

LogvinovLeon:
Now those preflight requests are not cached and executed before each call.
This causes performance problems if the DApp needs to fetch a lot of data.
This pool-request aims to fix that by introducing a 5 min caching for preflight CORS requests.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -32,7 +32,8 @@ module.exports = function(options) {
       var headers = {
         "Access-Control-Allow-Headers": "Origin, X-Requested-With, Content-Type, Accept",
         "Access-Control-Allow-Origin": "*",
-        "Access-Control-Allow-Methods": "*"
+        "Access-Control-Allow-Methods": "*",
+        "Access-Control-Max-Age": "300"
       };
 
       switch (method) {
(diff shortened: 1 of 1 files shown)


fanatid:
Sorry, I'm not familiar with  testrpc, but why not cache requests on client side?

LogvinovLeon:
The requests would be cached client-side. The header is just a hint to a browser for how long can it cache those requests.
For now my network tab looks like:
* OPTIONS
* get some actual data
* OPTIONS
* get some actual data
* OPTIONS
* get some actual data
* ....
* OPTIONS
* get some actual data

Those [preflight requests](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request) are not neccesary before each request if your server doesn't have dynamic CORS policy.

I would like it to look like:
* OPTIONS
* get some actual data
* get some actual data
* get some actual data
* ....
* get some actual data

benjamincburns:
Sorry, but the logic you're patching has moved to [trufflesuite/ganache-core](https://github.com/trufflesuite/ganache-core). If you still feel this change is necessary, can you please reopen this PR against the `develop` branch in that repo?

PR #301, opened 2017-04-25
issue 300: adding missing functions so testrpc can work with redux-devtools

sogoiii:
Request add missing rpc calls so redux-devtools works. 

1. eth_protocolVersion
2. personal_listAccounts
3. bzz_hive
4. bzz_info
5. ssh_version

Nothing is using those methods currently. 
@tcoulter

### DIFF
--- a/lib/subproviders/geth_api_double.js
+++ b/lib/subproviders/geth_api_double.js
@@ -277,6 +277,26 @@ GethApiDouble.prototype.eth_uninstallFilter = function(filter_id, callback) {
   callback(null, true);
 };
 
+GethApiDouble.prototype.eth_protocolVersion = function(callback) {
+  callback(null, "63");
+};
+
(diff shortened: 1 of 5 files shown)


tcoulter:
Hi there, thanks for the PR!

Although most of these requests are pretty static, would you mind submitting tests with them as well? If you had to submit one, the one I'd choose is `personal_listAccounts`, since the data it returns is generated and should equal what's returned by `eth_getAccounts` (I believe)

sogoiii:
@tcoulter Coolio, I added tests! 

I had to skip the bzz.getInfo and bzz.getHive because they are not in web3 version 0.16.0. 
I did try to update to 0.18.4, but other tests started to fail so reverted and skipped them.

tcoulter:
This has been merged into `ganache-core` (and I took care of conflicts). Cheers!

PR #305, opened 2017-04-28
Expose setTime to API

danielyaa5:
I was having difficulty reseting time back to original time for my unit tests after making the `evm_increaseTime` call to change the time. I saw the setTime function wasn't exposed and was so easy to expose that I decided to make a PR before even getting feedback.


axic:
I'd suggest to use the same API as `cpp-ethereum` (see https://github.com/ethereum/interfaces/issues/4). Perhaps call it `evm_modifyTimestamp` if it is kept under the `evm` namespace.

bashalex:
@axic I'd like to open another PR with this functionality. This method would be very useful for unit tests.
Do you want `evm_modifyTimestamp` to exist along with `evm_increaseTime` or replace it?

jakub-wojciechowski:
Hi @danielyaa5, @axic, @bashalex, 
I totally agree that we need such a feature. 

I've just pushed a [PR](https://github.com/trufflesuite/ganache-core/pull/13) to [ganache-core](https://github.com/trufflesuite/ganache-core) with a proposal of an atomic `evm_mineAtTimestamp` api function. Let me know if you have any comments.

benjamincburns:
Closing this one, as the logic has moved to ganache-core, as discussed above.

PR #309, opened 2017-05-15
Web3-provider-engine update

Mikefluff:
testrpc started to work incorrect with new truffle 3.2.2 compiler. The problem has been resolved after testrpc web3-provider-engine. I first tried to use version 12.0.5 but it looks like additional fixes should be done cause 12 branch uses external block-tracker. Will try to fix this in next pull request.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,8 +32,7 @@
     "solc": "0.4.6",
     "temp": "^0.8.3",
     "tmp": "0.0.31",
-    "web3": "~0.16.0",
-    "web3-provider-engine": "~8.1.0",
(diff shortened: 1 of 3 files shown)


benjamincburns:
I'm uncertain as to whether this change is still necessary. I think it's not? Either way, the logic you're patching has since moved to [trufflesuite/ganache-core](https://github.com/trufflesuite/ganache-core).

Sorry to close this straight off, but please reopen the PR against the `develop` branch in that repo if you think this is still necessary. I'll be sure to look into it quickly.

PR #314, opened 2017-05-30
Increase entropy when creating bip39 mnemonic

kevsul:
In the latest version of bip39 (2.3.1), entropyToMnemonic() requires at least 32 bytes of entropy


tcoulter:
This is awesome, thanks @kevsul. I've included this in the `release` branch of trufflesuite/ganache-core, and this will go out with the next release of the testrpc. Thanks!

PR #334, opened 2017-06-24
Fixed broken web3 dependency

dylanseago:
Upgraded web3 to ^0.19.1 and web3-provider-engine to ~13.0.0 to fix https://github.com/ethereumjs/testrpc/issues/332

Tests still need to be fixed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ethereumjs-testrpc",
-  "version": "3.0.5",
+  "version": "3.0.6",
   "bin": {
     "testrpc": "./bin/testrpc"
   },
--- a/package.json
+++ b/package.json
@@ -28,8 +28,8 @@
     "seedrandom": "~2.4.2",
     "shelljs": "~0.6.0",
     "solc": "0.4.6",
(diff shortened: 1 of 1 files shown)


historiographist:
thanks for this, but it seems this is failing tests for a reason. even with these upgraded deps, i'm still getting errors when i run contracts, a la "TypeError: self.engine._fetchBlock is not a function" (as in the failed tests)

dylanseago:
@drkhaira Thanks for noting that. There are definitely things that need to be fixed to support the upgrade, this PR should just be used as a starting point for resolving the issues caused by https://github.com/ethereum/web3.js/issues/904

tcoulter:
Hi there. Thanks for the PR.

This PR does two things: Upgrades web3 *and* upgrades the provider engine. Upgrading the provider engine should be a separate issue tackled at a separate time (although for the error you mentioned, bignumber.js, updating the provider engine might be required). I've updated web3 on the release branch of trufflesuite/ganache-core, which is now the core of the TestRPC. We'll update the provider engine in a future ticket. 

Thanks!

PR #342, opened 2017-07-06
Add feature which creates a json file with private keys in the workin…

kangarang:
…g directory of where testrpc is executed

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -9,6 +9,25 @@ var URL = require("url");
 var Web3 = require("web3");
 var web3 = new Web3(); // Used only for its BigNumber library.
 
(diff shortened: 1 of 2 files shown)


seunlanlege:
why was this closed? This featured would've made my life easier.

mikeseese:
@seunlanlege I'm not sure why it was originally closed by the author. However looking at these changes now, we wouldn't take it due to bluebird promises (instead of ES6 promises), the fixed amount of keys, and no option to not output the file (or to output it somewhere else).

Further, the private keys are being output on stdout.

If you'd like this feature still, please open a new issue and we can discuss further. Thanks!

PR #343, opened 2017-07-06
add the correct entry point for docker

ezynda3:
This fixes the current Dockerfile build by specifying the correct path to the testrpc CLI.


tcoulter:
Thanks!

Issue #349, opened 2017-07-17
Docker image: Could not find module '/src/bin/testrpc'


mikeshultz:
## Expected Behavior
Building docker container from source and running it should result in a working testrpc server.

## Current Behavior
Container crashes with the following in `docker logs`

    $ docker logs b5e74bb1f422b0656216523b44da9f72856c24f9ac29955525711e7ecdf1bc4c
    module.js:471
        throw err;
        ^

    Error: Cannot find module '/src/bin/testrpc'
        at Function.Module._resolveFilename (module.js:469:15)
        at Function.Module._load (module.js:417:25)
        at Module.runMain (module.js:604:10)
        at run (bootstrap_node.js:394:7)
        at startup (bootstrap_node.js:149:9)
        at bootstrap_node.js:509:3
    module.js:471
        throw err;
        ^

    Error: Cannot find module '/src/bin/testrpc'
        at Function.Module._resolveFilename (module.js:469:15)
        at Function.Module._load (module.js:417:25)
        at Module.runMain (module.js:604:10)
        at run (bootstrap_node.js:394:7)
        at startup (bootstrap_node.js:149:9)
        at bootstrap_node.js:509:3

## Steps to Reproduce (for bugs)
1. `git clone https://github.com/ethereumjs/testrpc.git && cd testrpc`
2. `docker build -t ethereumjs/testrpc .`
3. `docker run -d -p 8545:8545 ethereumjs/testrpc -a 10 --debug`
4.`docker logs [container id]`

## Your Environment
Operating System and version: Ubuntu 17.04

noisy:
I can confirm that behavior. From what I see `bin/testrpc` was removed in 464039b32396e6fac406ecc087d9900e7d47268d by @tcoulter

mikeshultz:
Tested and it works.  Thanks

PR #352, opened 2017-07-19
Fix for #349. Replaced ./bin/testrpc by ./cli.js after file renaming in 464039b32396e6fac406ecc087d9900e7d47268d

noisy:
Fix for #349.

testrpc file was renamed in commit: 464039b32396e6fac406ecc087d9900e7d47268d, but Dockerfile wasn't updated.


tcoulter:
Thanks!

Issue #365, opened 2017-08-05
Server won't close because of keep alive connections


krzkaczor:
## Expected Behavior
Hitting ctrl+c (SIGINT) should always close the server. 

## Current Behavior
If there are connections kept alive (for example meta mask plugged into the local network) server will *never* stop. It's [expected behavior](https://github.com/nodejs/node-v0.x-archive/issues/9066) for` httpServer.close` but definitely not expected behavior for the user. 

## Possible Solution
Use `server-destroy` package which internally stores information about all connections and closes them before closing server. I will post a PR in a sec.

## Steps to Reproduce (for bugs)
1. Start RPC
2. Set up `metamask` to connect to local network.
3. Try closing the server.
4. NO PROFIT :(

## Your Environment
* Version used: 4.0.1
* Environment name and version (e.g. PHP 5.4 on nginx 1.9.1): node v8.1.4
* Operating System and version: MacOS

mattgstevens:
I also have this problem when using [concurrently](https://github.com/kimmobrunfeldt/concurrently) and the --kill-others flag.

Current behaviour of testrpc means it may not shutdown on SIGTERM which makes local development less nice, since you will need to use a command like `pkill -f testpc` to stop the testrpc

step21:
Seconded. How to kill it?

szerintedmi:
I've been struggling with this issue for a while and I found the following workaround which works with high enough nodejs versions:
Ctrl-Z 
then `kill %1`

the trick that it's only working with node > 8.3.0 , I've no clue why :)
I played a bit and 8.5.0 I've found the best version to use so far. 
With 8.4.0 it seemed to have worse performance and with new versions it regularly crashes.

PR #366, opened 2017-08-05
Force close server, even if there are open connections

krzkaczor:
Closes https://github.com/ethereumjs/testrpc/issues/365

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,8 @@
   },
   "devDependencies": {
     "ganache-core": "1.1.3",
-    "yargs": "^7.0.2"
+    "yargs": "^7.0.2",
+    "server-destroy": "^1.0.1"
   },
   "repository": {
     "type": "git",
--- a/cli.js
+++ b/cli.js
@@ -3,6 +3,7 @@
 // https://github.com/yargs/yargs/issues/781
 var yargs = require('yargs/yargs');
 var Ganache = require("ganache-core");
+var enableDestroy = require("server-destroy");
 var pkg = require("./package.json");
 var corepkg = require("./node_modules/ganache-core/package.json");
 var URL = require("url");
@@ -99,6 +100,7 @@ if (options.fork) {
 }
 
 var server = Ganache.server(options);
+enableDestroy(server);
 
 //console.log("Ganache CLI v" + pkg.version);
 console.log("EthereumJS TestRPC v" + pkg.version + " (ganache-core: " + corepkg.version + ")");
@@ -187,7 +189,7 @@ if (process.platform === "win32") {
 
 process.on("SIGINT", function () {
   // graceful shutdown
-  server.close(function(err) {
+  server.destroy(function(err) {
     if (err) {
       console.log(err.stack || err);
     }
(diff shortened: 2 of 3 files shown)


mattgstevens:
@krzkaczor think that the merge target needs to be `ethereumjs:develop` instead of `master` branch

mattgstevens:
@krzkaczor ~~im going to open another PR that targets the correct branch since you havn't update the target branch in some weeks. would very much like to have this behaviour in master :)~~

im unsure of how things get merged in this project as the TravisCI will always error, and the develop branch is certainly stale.

@tcoulter any reason to not merge this in / how can we get this PR into an acceptable state?

package.json
```diff
@@ -27,6 +27,7 @@
     "url": "https://github.com/ethereumjs/testrpc"
   },
   "dependencies": {
+    "server-destroy": "^1.0.1",
```
moodysalem: i think you want this to be a dev dependency since the distribution is built with webpack

mattgstevens: good point, will make a new PR tomorrow if @krzkaczor continues to be quiet on this; it is quite annoying for me in dev workflow

krzkaczor:
Hey guys @moodysalem @mattgstevens 
Sorry for delays - I am super busy with work and stuff. I was using my fork for some time and it for sure improves the situation but I noticed that meta mask still can behave weird. Like it's not re-establishing a new connection and fails silently. I am not sure if other ppl experience similar problems. I think we should leave a note on readme at least.

mattgstevens:
@krzkaczor no worries; open source takes time :)

Think if you update this PR to move `server-destroy` into devDependencies as @moodysalem suggested it will be an easy yes to merge.

Not sure about the MetaMask issues you are mentioning, and seems like a separate issue. FWIW the dev environment I'm working in uses [concurrently](https://github.com/kimmobrunfeldt/concurrently) to start `testrpc` as one process and whatever you have for dev serving (webpack / express / ...) as another process. When the source is changing the watcher in the second process can do its thing, and the `testrpc` porcess stays alive.

krzkaczor:
@mattgstevens  i am on it riight now. But I can't install project again with `yarn` because of some weird `node-gyp` problems. I think yarn 1.x broke it Oo

krzkaczor:
@mattgstevens done. I rebased with master but it seems like its in corrupted state - doing yarn on master generates massive diff already. 🙄

moodysalem:
that's a massive diff for adding a package that has no dependencies, maybe you should just start from master?

krzkaczor:
@moodysalem the same happens on master - that's the thing. Just run `yarn` on master. It generates changes in few files.

benjamincburns:
@krzkaczor the real problem was that we were committing generated build artifacts. I've since updated the `develop` branch to drop these. I think if you update this PR to drop these files from your commit, we'll be in good shape.

That said, I'm not well familiar with the dependency you're introducing here. What is it offering us? Should this be something which is actually handled in [trufflesuite/ganache-core](https://github.com/trufflesuite/ganache-core)?

step21:
I would really like to see this. Is it ok to resubmit this, as I cannot update this PR obviously. I think server destroy is an ok dependency, it might also be possible to handle it in ganache-core but I do not really know ganache-core so not sure.

szerintedmi:
this is a much waited pull request :) until it's merged I've found this workaround: https://github.com/trufflesuite/ganache-cli/issues/365#issuecomment-357098984

benjamincburns:
I've fixed this in the `develop` branch of ganache-core.

PR #368, opened 2017-08-13
Corrected Gas Limit in README

0xcaff:
It looks like it was changed in fc0fb4cc728c1eb7df1745120151f9ef8a737584 without the README being updated.


benjamincburns:
also fixed up the default in the CLI usage

PR #392, opened 2017-09-25
Double quotes around mnemonic

ghost:
I was struggling to get this to work until I read in one of the issues that it needs double quotes.


benjamincburns:
It's a nice note, but I'd rather leave it off for two reasons.

1) The need to quote or escape items with spaces is well understood in CLI environments, and the requirement to do so is imposed by your shell, not by TestRPC. I'd rather not delay a CLI newbie's education by confusing the source of this requirement.
2) It's pedantic, but while it's true that double quoting will work fine for all the shells I'm aware of, it's not the only quoting mechanism, and it's possible that another may suit better. I'd rather not make the recommendation.

PR #400, opened 2017-10-10
Fix README argv desc

XatMassacrE:
In the READEME file(42 line):

* `-i` or `--network-id`: Specify the network id the TestRPC will ...

But I found the 76 line of `cli.js` file:
```
network_id: argv.i || argv.networkId,
```

Although `--network-id` still work(yargs parse?), but it's confused. And i think it's better to use `--networkId` and `--networkId` can stay the same with `gasPrice`, `gasLimit` and the later **options desc**.


benjamincburns:
Sorry it took so long to accept this!

PR #406, opened 2017-10-23
Add ability to save accounts and private keys on startup

mattkanwisher:
Add the ability to dump accounts and private keys to disk. So other programs can programmatically read and use them

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@ Both `.provider()` and `.server()` take a single object which allows you to spec
 * `"locked"`: `boolean` - whether or not accounts are locked by default.
 * `"unlocked_accounts"`: `Array` - array of addresses or address indexes specifying which accounts should be unlocked.
 * `"db_path"`: `String` - Specify a path to a directory to save the chain database. If a database already exists, the TestRPC will initialize that chain instead of creating a new one.
+* `"account_keys_path"`: `String` - Specifies a file to save accounts and private keys to, for testing.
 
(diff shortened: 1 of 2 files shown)


benjamincburns:
Thanks for the change. I haven't tested it yet, but as long as there's no race condition on the creation of `accounts`, I'd think it should be fine. My only other concern is that I'm wondering how well it fits in with #23 -- in that case you'd be able to specify the accounts, meaning you should already know your private key. I suppose there's no harm in dumping it back out again in that case, however.

So I guess the only question I have @mattkanwisher is whether you would've submitted this PR even if #23 (genesis.json support) were closed/implemented today?

mattkanwisher:
I think this is slightly different usecase then #23 cause I want testrpc to create a bunch of test accounts, then I have another program that wants to be able to use that blockchain. If you have to supply the genesis file, the other program has to know how to correctly create accounts and private keys, which testrpc already does. I see this feature tying into the Database saving feature, cause currently every time you launch testrpc even if you persist the database it recreates the users, it would be nice if it reused the users, and was able to use persistent state.

mattkanwisher:
The other thing I could do, is modify my json file so it generates a portion of this genesis file? But it would need to somehow be able to tie into the saved database with the --db_path option

benjamincburns:
Going to accept this as-is, as I see the use for it now - when we get the genesis file underway, we'll add a separate option to save out a genesis file for repeatability.

PR #413, opened 2017-11-03
Fix the docker builds for v6 releases

roderik:
Since removing the build assets from the repo, the docker builds no longer work. This PR fixes that and upgrades tot he latest node 8 LTS release


benjamincburns:
Thanks for this, @roderik!

PR #481, opened 2018-02-14
Shed cli logic

benjamincburns:
Don't merge until after trufflesuite/ganache-core#67 is merged.

Also I will likely need to update `package-lock.json` again when that happens.


benjamincburns:
@davidmurdoch is doing some work which invalidates the need for this

PR #483, opened 2018-02-18
fix: All instances of "blocktime" => "blockTime"

devonzuegel:
The documentation points to using the `--blocktime` flag, but it caused ganache to crash. It works as expected when I changed it to `--blockTime`, so wanted to update the docs accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Options:
 
 * `-a` or `--accounts`: Specify the number of accounts to generate at startup.
 * `-e` or `--defaultBalanceEther`: Amount of ether to assign each test account. Default is 100.
-* `-b` or `--blocktime`: Specify blocktime in seconds for automatic mining. Default is 0 and no auto-mining.
+* `-b` or `--blockTime`: Specify blockTime in seconds for automatic mining. Default is 0 and no auto-mining.
 * `-d` or `--deterministic`: Generate deterministic addresses based on a pre-defined mnemonic.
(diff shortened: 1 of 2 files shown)


benjamincburns:
Thanks for this @devonzuegel!

FWIW, kabob-case (i.e. `--block-time`) should also work. I think `--blocktime` was from the args library we used before we introduced `yargs`. That is, we were specifying the argument as `blockTime` but that lib flattened the casing to `blocktime`, and we forgot to update our docs when things changed.

benjamincburns:
Also just as a heads up, using the `--blockTime` flag at all is somewhat discouraged unless you have tests which require a specific mining interval. If you don't specify this flag, ganache will instantly mine a new block for every transaction, and do nothing if there are no pending transactions. Instamining (what we call the default mode of operation when no `--blockTime` flag is specified) will run your tests much faster than interval mining.

devonzuegel:
Sweet! 

Yep, we usually don't have `blockTime` on for the project, but we needed it for a particular case to simulate polling our votes. Thanks for the heads up. :)

ngan:
Not sure if it was because of this PR...but setting "--blockTime 1" has no effect. Blocks were not getting mined every 1 second.

Just FYI, we'd use this feature in development. We look for 5 block confirmations before considering a block final in our app. This allows us to emulate production environment where blocks are continuously being mined.

kingjerod:
@ngan I opened an issue for it: https://github.com/trufflesuite/ganache-cli/issues/506

PR #519, opened 2018-05-09
Update the blockTime option description

etnann:
I expanded and correct the current description which is currently ambiguous (why does 0 seconds mean? If each transaction is mined in 0 seconds (instantaneously), why then the documentation states `no auto-mining.`)

I took part of the description from  https://github.com/trufflesuite/ganache-cli/pull/483#issuecomment-367553231


mikeseese (approved):
Looks good to me!

Thanks for contributing!

PR #537, opened 2018-06-15
Move webpack-cli to devDependencies

### LINKED ISSUE #522: Remove webpack-cli from production dependencies
Please correct me if I'm wrong but I think `webpack-cli` should be in `devDependencies`. This is resulting in a lot of unnecessary installed packages for projects that depend on Truffle.

frangio:
Fixes #522.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -48,6 +48,7 @@
       "version": "6.2.0",
       "resolved": "https://registry.npmjs.org/ajv/-/ajv-6.2.0.tgz",
       "integrity": "sha1-r6wpW7qgFSRJ5SJ0LkVHwa6TKNI=",
+      "dev": true,
       "requires": {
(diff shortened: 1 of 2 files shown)


mikeseese:
For the record, it appears that `fsevents` was added in the `package-lock.json` It appears to be an `optionalDependency` for `chokidar`. I don't see any harm in it being added however.

@frangio Out of curiosity, are you on a Mac? I would guess why the FSEvents was installed:
> Native access to OS X FSEvents in Node.js

mikeseese (approved):
LGTM! I'm going to wait for @benjamincburns to comment on our slack to see if he has any objections before merging

frangio:
I'm on Linux so I don't know why that happened. I've lost hope on trying to understand the lockfile changes done by npm.

PR #539, opened 2018-06-21
Fix broken --host option due to being overridden by help option

FunkyM:
The help option was defined to use the 'h' shorthand option.
However, this was overriding the previous host option which also used the 'h' shorthand option.
Thus running ganache-cli with the '-h' or even '--host' option did not work and showed the help output.
Now only the '--help' and '-?' options trigger the help output which is the expected behaviour.

I'd be happy if you could do a minor release if this is fixed/merged as now any setup that uses the host option is broken... other than that, great work on the very helpful ganache-cli tool!

### DIFF
--- a/args.js
+++ b/args.js
@@ -138,8 +138,8 @@ module.exports = exports = function(yargs, version) {
       type: 'boolean',
       default: false
     })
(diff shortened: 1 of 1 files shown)


mikeseese (approved):
LGTM!

Thanks a lot for contributing!

PR #548, opened 2018-06-29
Fix faulty `--hostname` CLI option

dragossutu:
Since the _-h_ option was fixed, `argv.h` is mostly truthy, it either has the value passed through CLI or the default value _127.0.0.1_.
Thus `|| argv.hostname || "127.0.0.1"` is dead code and _--hostname_ option cannot be used to set the hostname anymore.

Removed dead code, use the default value set in yargs.
Added _--hostname_ option in yargs.

### DIFF
--- a/args.js
+++ b/args.js
@@ -9,7 +9,7 @@ module.exports = exports = function(yargs, version) {
     })
     .option('h', {
       group: 'Network:',
-      alias: 'host',
+      alias: ['host', 'hostname'],
       default: '127.0.0.1',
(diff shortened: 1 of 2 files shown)


mikeseese (approved):
Thanks @drgsutu for reporting/fixing this!

It looks like we'll need to cleanup some of the other dead code that became redundant after args.js was changed a little back.

This all looks good, so I'll merge it and we'll do a clean up of other dead code  on our own just to make sure this fix gets in before the next release.

Thanks again!

PR #550, opened 2018-07-02
Add flag for unlimited contract size

mikeseese:
This PR adds the flag for `allowUnlimitedContractSize` which was added to `ganache-core` in https://github.com/trufflesuite/ganache-core/pull/126


benjamincburns (commented):
👍

PR #560, opened 2018-08-10
Remove 'wei' units from block gas limit parameter

4c656554:
As I understand it, the block gas limit, `gasLimit` is in terms of the maximum EVM gas used, as opposed to the amount of Ether spent for each gas unit.    At present, the `-l` /`gasLimit` parameter is noted as being in units of Wei. The suggested minor update removes the units to avoid confusion for new users.  Thanks to the devs and maintainers for this useful project.


mikeseese (approved):
You are 100% correct. I was dumb to put that in; my mind must have skipped a beat when I was thinking that.

Thanks for catching this and taking the time to make the PR!

PR #563, opened 2018-08-15
Use Docker multistage builds for reduced imagesize

thomaseizinger:
This drastically reduces the size of the resulting docker image.

The image with the `multi-stage-build` tag is the new one (73MB uncompressed) whereas the other ones are downloaded from DockerHub (573MB).

image

Is there a way how I can run tests against the new image to see if it still works correctly?
I did a manual smoke test and it seems to be working fine!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM mhart/alpine-node:8
+FROM mhart/alpine-node:8 as builder
 
 RUN apk add --no-cache make gcc g++ python git bash
 COPY package.json /src/package.json
@@ -8,8 +8,14 @@ RUN npm install
 ADD . .
 RUN npm run build
 
+FROM mhart/alpine-node:8 as runtime
+
+WORKDIR /app
+COPY --from=builder "/src/build/cli.node.js" "./cli.node.js"
+COPY --from=builder "/src/build/cli.node.js.map" "./cli.node.js.map"
+
 ENV DOCKER true
 
 EXPOSE 8545
 
-ENTRYPOINT ["node", "./build/cli.node.js"]
+ENTRYPOINT ["node", "/app/cli.node.js"]


davidmurdoch:
Thanks @thomaseizinger! This looks very promising! I'm not a docker guru myself, so I'll tag in @benjamincburns or @seesemichaelj to get their feedback on this.

@seesemichaelj or @benjamincburns We should probably run our tests within the docker build, which looks like it'd be simple to configure with a multi-stage build, per: https://codefresh.io/docker-tutorial/node_docker_multistage/#creatingamulti-stagedockerfilefornode.jsapplication

thomaseizinger:
Running them within the Docker is not as useful IMO because there are no
source code changes compared to outside of Docker. Integration-tests
against the HTTP API from the outside would shed light onto whether the
system behaves correctly.

benjamincburns:
I like this idea, haven't reviewed implementation, however -- I'll leave that to you @davidmurdoch!

mikeseese:
I actually already reviewed the implementation; this looks great! I was in the middle of testing when I discovered an issue with our latest release that I had to handle :sweat_smile: 

Looking at this now

mikeseese (changes requested):
I tested this (with the latest develop changes), and it worked as expected.

It should be noted that during release of Docker images, that two images are built, only the second one (smaller one) will need to be published.

One requested change however: you should also copy `/src/build/cli.node.js.map` over to the `/app` folder. This is our source mapping file.

thomaseizinger:
@seesemichaelj Just incorporated the request changes!

mikeseese (approved):
LGTM! Thanks again!

thomaseizinger:
Thanks for merging this so quickly!

Any chance that this is backported to the v6.1.x release? The repository is already tracking the 7.x version right?

mikeseese:
The 7.x series was not continued past `v7.0.0-beta.0`. I'm guessing @benjamincburns changed his mind and went to 6.1.0 instead.

The current series is the 6.1.x series. This PR will go into effect in 6.1.9, 6.2.0, or 7.0.0 (whatever is the next release, more than likely 6.1.9).

Perhaps we should look into removing the `v7.0.0-beta.0` image from docker hub.

thomaseizinger:
Oh. good to know!

Are there already plans for the next release?

mikeseese:
Unfortunately no, as it's needed. We've been trying to do it regularly (seems ~2 weeks). Sorry I don't have more details!

PR #577, opened 2018-10-07
Update README.md

TheAlgo:


davidmurdoch:
Thanks for the PR, @TheAlgo. Why do you want to add the word "procedure" here?

PR #578, opened 2018-10-09
Allow silencing output

CruzMolina:
When running a large volume of tests/transactions against the local ganache-cli EVM (especially when using a custom config with a large amount of account Ether balance), it would be nice to be able to silence any log output in order to increase performance.

## Steps to Test

In one terminal window:
`ganache-cli -q` or `ganache-cli --quiet`

In another terminal window:
`cd myTruffleProjectWithTests`
`truffle test`

## Expected Outcome

Truffle tests run and ganache-cli doesn't output anything after the initial start and below the following line:
`Listening on 127.0.0.1:8545`


davidmurdoch:
Works wonderfully. Thanks @CruzMolina!

PR #616, opened 2019-01-25
Update README.md

mikeislearning:
Making clear that `balance` is the correct word to use when constructing the accounts objects


davidmurdoch (approved):
LGTM!

PR #659, opened 2019-06-24
Display Available Account addresses as EIP-55 checksum addresses when ganache-cli starts

djuanit0x:


davidmurdoch:
Thoughts on including [EIP-1191](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1191.md)'s algorithm here?

https://github.com/ethereumjs/ethereumjs-util/blob/3b1085059194b02354177d334f89cd82a5187883/src/account.ts#L42-L59

Alternatively, we could just use the `ethereumjs-util` itself, which already implements the algorithm and will make it easy for me to add in EIP-1191-based checksumming in an upcoming version.

djuanit0x:
Adding EIP-1191's toCheckSumAddress from `ethereumjs-util`, would be great, but should we do that in this PR or in another PR?

davidmurdoch:
You can just commit to this PR to update it. No need to create a new PR.

davidmurdoch:
Changes look great! I'll test things soon and can hopefully get this out in the next beta release this week!

PR #704, opened 2019-12-24
Add hd_path to cli

GregTheGreek:
This adds the ability to change the base path used to derive from an HD wallet.

This pr pairs with [a pr](https://github.com/trufflesuite/ganache-core/pull/519) in ganache-core.

### DIFF
--- a/args.js
+++ b/args.js
@@ -97,8 +97,9 @@ module.exports = exports = function(yargs, version, isDocker) {
       demandOption: false
     })
     .option('hdPath', {
-        group: 'Chain:',
-        describe: 'The base path used to derive the accounts from.',
+        group: 'Accounts:',
+        alias: 'hd_path',
+        describe: `The hierarchical deterministic path to use when generating accounts. Default: "m/44'/60'/0'/0/"`,
         type: 'string',
         demandOption: false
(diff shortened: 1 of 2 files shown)


args.js
```diff
@@ -96,6 +96,12 @@ module.exports = exports = function(yargs, version, isDocker) {
       conflicts: 'd',
       demandOption: false
     })
+    .option('hdPath', {
+        group: 'Chain:',
+        describe: 'The base path used to derive the accounts from.',
+        type: 'string',
+        demandOption: false
+    })
```
davidmurdoch, line 105: Looks great! Can you add the alias `hd_path`? Also, the `group` should be `Accounts:` and update the `describe` as well?.

```
.option('hdPath', {
  group: 'Accounts:',
  alias: 'hd_path',
  describe: 'The hierarchical deterministic path to use when generating accounts. Default: "m/44'/60'/0'/0/"',
  type: 'string',
  demandOption: false
})
```

Are you able to make this change? If not I can do it.

Thanks!

davidmurdoch (changes requested):

davidmurdoch (approved):

Issue #733, opened 2020-04-27
Investigate implementing a daemon mode


davidmurdoch:
For technical reasons, ganache-core cannot queue HTTP requests immediately on initialization -- the underlining HTTP server needs time to start up. This makes it difficult to run automated test suites using `ganache-cli`. Example:

```json
  "test": "npx ganache-cli && npx truffle test ... "
```

will _not_ wait for ganache-cli's http server to be ready before sending requests.

The current workaround is to do things like `(sleep 2 && npx truffle test) & npx ganache-cli`, which is just horrible horrible horrible.

If we add a daemon mode, i.e., `ganache-cli --daemon` that spawns a ganache-server _in the background_, waits for it to be ready for HTTP requests, then exits, we could do things like:

```json
  "test": "npx ganache-cli && npx truffle test ... "
```

We'd need some way of killing running daemon's as well, of course. Something like `npx ganache-cli --daemon-stop ID` and/or `npx ganache-cli --daemon-stop-all`.

anthanh:
maybe related to this other issue? https://github.com/trufflesuite/ganache-cli/issues/675

PR #735, opened 2020-04-29
Feat/deasync startup

nicholasjpaterno:
Addresses #733

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,7 @@ Both `.provider()` and `.server()` take a single object which allows you to spec
 * `"gasLimit"`: `String::hex | number` Sets the block gas limit. Must be specified as a `hex` string or `number`. Defaults to `"0x6691b7"`.
 * `"callGasLimit"`: `number` Sets the transaction gas limit for `eth_call` and `eth_estimateGas` calls. Must be specified as a `hex` string. Defaults to `"0x1fffffffffffff"` (`Number.MAX_SAFE_INTEGER`).
 * `"keepAliveTimeout"`:  `number` If using `.server()` - Sets the HTTP server's `keepAliveTimeout` in milliseconds. See the [NodeJS HTTP docs](https://nodejs.org/api/http.html#http_server_keepalivetimeout) for details. `5000` by default.
+* `"deasync"`:  `boolean` Synchronizes ganache server startup. Useful in certain scenarios - see ganache-cli issue [#733](https://github.com/trufflesuite/ganache-cli/issues/733).
 
 ## Implemented Methods
 
--- a/cli.js
+++ b/cli.js
@@ -6,6 +6,7 @@ require('source-map-support').install();
 var yargs = require("yargs");
 var pkg = require("./package.json");
 var {toChecksumAddress, BN} = require("ethereumjs-util");
+var deasync = require('deasync');
 var ganache;
 try {
   ganache = require("./lib");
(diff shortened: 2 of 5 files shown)


cli.js
```diff
@@ -190,7 +190,10 @@ server.listen(options.port, options.hostname, function(err, result) {
 
   console.log("");
   console.log("Listening on " + options.hostname + ":" + options.port);
-});
+} catch (error) {
+  console.log(error);
```
davidmurdoch: What do you think about updating this to `console.error`? Hmmmm. wonder if that'd break anyone's build if they check for errors on stdout instead of stderr. :-/

cli.js
```diff
@@ -190,7 +190,10 @@ server.listen(options.port, options.hostname, function(err, result) {
...
+  return;
+}
 
 process.on('uncaughtException', function(e) {
```
davidmurdoch: Do you think these `process.on(...` handlers should go before the listen call now? Or doesn't it not matter since it's synchronous-ish now?

cli.js
```diff
@@ -6,6 +6,7 @@ require('source-map-support').install();
 var yargs = require("yargs");
 var pkg = require("./package.json");
 var {toChecksumAddress, BN} = require("ethereumjs-util");
+var deasync = require('deasync');
```
davidmurdoch: <pedantic>Use double quotes here.

davidmurdoch (changes requested):

cli.js
nicholasjpaterno: Definitely!  I actually was thinking about that as I was trying to fall asleep.

cli.js
nicholasjpaterno: That was my concern as well.

davidmurdoch:
I've updated this PR, but I'm still not confident in its stability. Might need to beta this one for a few weeks.

nicholasjpaterno (commented):
LGTM

cli.js
```diff
@@ -138,86 +143,100 @@ process.on("SIGINT", function () {
...
-  console.log("Time:        " + (state.blockchain.startTime || new Date()).toString());
+try {
+  if (deasync) {
+    const listen = deasync(server.listen);
+    const result = listen(options.port, options.hostname);
+    startGanache(null, result);
+  } else {
+    server.listen(options.port, options.hostname, startGanache);
+  }
+} catch (error) {
```
davidmurdoch: The `try`/`catch` here may cause a failure on Windows which is why I removed it in the previous commit.

README.md
```diff
@@ -168,6 +168,7 @@ Both `.provider()` and `.server()` take a single object which allows you to spec
 * `"gasLimit"`: `String::hex | number` Sets the block gas limit. Must be specified as a `hex` string or `number`. Defaults to `"0x6691b7"`.
 * `"callGasLimit"`: `number` Sets the transaction gas limit for `eth_call` and `eth_estimateGas` calls. Must be specified as a `hex` string. Defaults to `"0x1fffffffffffff"` (`Number.MAX_SAFE_INTEGER`).
 * `"keepAliveTimeout"`:  `number` If using `.server()` - Sets the HTTP server's `keepAliveTimeout` in milliseconds. See the [NodeJS HTTP docs](https://nodejs.org/api/http.html#http_server_keepalivetimeout) for details. `5000` by default.
+* `"deasync"`:  `boolean` Synchronizes ganache server startup. Useful in certain scenarios - see ganache-cli issue [#733](https://github.com/trufflesuite/ganache-cli/issues/733).
```
davidmurdoch, line 171: Thoughts on changing to `--experimental-deasync` until we assess the viability?

davidmurdoch (changes requested):

README.md
nicholasjpaterno: I wonder if it makes sense to just put experimental in the description so we dont break anything when/if we remove `experimental`?

davidmurdoch: I was thinking we could always keep the `experimental` flag around as an alias.
Meh. Let's just ship it as is. :shipit:

davidmurdoch (approved):

PR #739, opened 2020-05-05
fix: Handle SIGINT, SIGTERM, & SIGHUP gracefully

### LINKED ISSUE #738: Non-zero exit code in response to SIGTERM
### bug

`ganache-cli` does respond to sigterm and shutsdown, but does so with a failure exit code.
since the shutdown happened cleanly, it should exit with a zero error code. this is useful in tests that rely on exit codes detect failures.

### reproduction

start ganache cli
```bash
npx ganache-cli
```
in another terminal
```
# get ganache pid via port usage
lsof -ti:8545
# send SIGTERM (15) to ganache PID
kill -15 $(lsof -ti:8545)
```
back in the other terminal
```
# get exit code of previous command
echo $?
143
```

nicholasjpaterno:
Only exit with code 0 when gracefully exiting.

Fixes #738

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -205,6 +205,9 @@ if (process.platform === "win32") {
   })
   .on("SIGINT", function () {
     process.emit("SIGINT");
+  })
+  .on("SIGHUP", function () {
+    process.emit("SIGHUP");
   });
 }
 
@@ -227,3 +230,4 @@ const closeHandler = function () {
 
 process.on("SIGINT", closeHandler);
 process.on("SIGTERM", closeHandler);
+process.on("SIGHUP", closeHandler);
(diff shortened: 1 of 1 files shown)


nicholasjpaterno (commented):
Related to #738

davidmurdoch (approved):
LGTM!

cli.js
```diff
@@ -205,15 +205,29 @@ if (process.platform === "win32") {
   })
   .on("SIGINT", function () {
     process.emit("SIGINT");
+  })
+  .on("SIGHUP", function () {
+    process.emit("SIGHUP");
```
davidmurdoch: Remove this. It isn't needed.

davidmurdoch (changes requested):

davidmurdoch (approved):

PR #743, opened 2020-05-13
build: add semantic-release

nicholasjpaterno:


.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,20 @@
+name: ganache-cli CD
+
+on:
+  push:
+    branches: [ master, beta ]
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
```
davidmurdoch: Change to a specific version so future upgrades don't break things.

davidmurdoch (changes requested):

PR #790, opened 2020-09-28
fix: remove unneeded COPY command from docker build

davidmurdoch:


tynes:
Was able to build the container locally after these changes

davidmurdoch:
@tynes i'll try to put out a beta tonight or tomorrow!

